PR #15, opened 2021-12-14
[castai-cluster-controller] add custom priority class

anjmao:


charts/castai-cluster-controller/templates/priority_class.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
```
varnastadeus, line 2: on older kops versions this was not working for agent deployment 😞

anjmao: Yes, I see it was added in `v1.14`. This one is actually quite important for spot only clusters.

Maybe we could have an object like:
```
priorityClass:
   enabled: true
```

And it will be disabled for older clusters. During phase2 generation we could probably check cluster k8s versions.

anjmao: We could try to use helm context variables to check which k8s version is used https://github.com/kubernetes/apimachinery/blob/master/pkg/version/types.go#L19-L32

PR #35, opened 2022-03-21
[castai-cluster-controller] add resource quota

anjmao:


charts/castai-cluster-controller/values.yaml
```diff
@@ -49,3 +49,7 @@ serviceAccount:
 
 aks:
   enabled: false
+
+priorityClass:
+  enabled: true
```
varnastadeus, line 54: will you set disabled for older versions?

varnastadeus (approved):

charts/castai-cluster-controller/values.yaml
anjmao: We will need to disable it only for old clusters during cluster-controller upgrade and customers who are using old k8s versions.

PR #36, opened 2022-03-25
[castai-cluster-controller] Granted all permissions within castai namespace

andrejatcastai:


charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: CAST AI cluster controller deployment chart.
 type: application
-version: 0.12.0
+version: 0.13.0
```
varnastadeus, line 5: Sorry, 0.13.0 is already deployed 

```suggestion
version: 0.14.0
```

charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: CAST AI cluster controller deployment chart.
 type: application
-version: 0.13.0
+version: 0.14.0
 appVersion: "v0.10.1"
```
varnastadeus, line 6: as you have changed cluster-controller, I would suggest to update appVersion in one go when it is released

charts/castai-cluster-controller/templates/rbac.yaml
```diff
@@ -73,19 +73,10 @@ metadata:
...
-  - apiGroups: [""]
-    resources: [ "serviceaccounts", "secrets", "configmaps" ]
-    verbs: [ "get", "list", "create", "update", "delete", "patch" ]
-  - apiGroups: [ "rbac.authorization.k8s.io" ]
-    resources: [ "roles" ]
-    verbs: [ "get", "list", "create", "update", "delete", "patch" ]
-  - apiGroups: [ "rbac.authorization.k8s.io" ]
-    resources: [ "rolebindings" ]
-    verbs: [ "get", "list", "patch", "update", "delete" ]
+  # controller can do anything in castai-agent namespace; this is required to fully manage castai components
```
anjmao, line 76: Need to test if this works with auto update.

charts/castai-cluster-controller/Chart.yaml
andrejatcastai: Got it, will changed app version here once it is released

anjmao (approved):

PR #38, opened 2022-03-29
[castai-cluster-controller] add pdb and leader election

anjmao:


charts/castai-cluster-controller/templates/rbac.yaml
```diff
@@ -73,7 +73,7 @@ metadata:
   labels:
     {{- include "castai-agent.labels" . | nindent 4 }}
 rules:
-  # controller can do anything in castai-agent namespace; this is required to fully manage castai components
+  # Controller can do anything in castai-agent namespace. This is required to fully manage castai components.
```
varnastadeus, line 76:
```suggestion
  # Controller has full access in castai-agent namespace. This is required to fully manage castai components.
```

varnastadeus (approved):

PR #61, opened 2022-05-09
[castai-agent] update agent version - allow to pass clusterID as env

aldor007:


charts/castai-agent/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-agent
 description: CAST AI agent deployment chart.
 type: application
-version: 0.20.1
-appVersion: "v0.25.1"
+version: 0.21.0
+appVersion: "v0.26.1"
```
dariuspie, line 6: If it was minor changes, increasing patch version would be enough. But in you are increasing minor, then reset patch to 0 consistently for both.

dariuspie (approved):

PR #78, opened 2022-05-20
[castai-cluster-controller] NOJIRA: add podantiaffinity to cluster-controller template

CastAIPhil:
Currently the cluster-controller pods can get scheduled on the same node which can cause FAILED mode if the node goes offline and the pods can't land somewhere else. In order to reduce the fragility of cluster-controller I have added a podAntiAffinity rule to force them onto separate nodes.

### DIFF
--- a/charts/castai-cluster-controller/templates/deployment.yaml
+++ b/charts/castai-cluster-controller/templates/deployment.yaml
@@ -42,6 +42,15 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       affinity:
+        podAntiAffinity:
+            requiredDuringSchedulingIgnoredDuringExecution:
+              - labelSelector:
+                  matchExpressions:
+                    - key: app.kubernetes.io/name
+                      operator: In
+                      values:
+                      - castai-cluster-controller
+                topologyKey: kubernetes.io/hostname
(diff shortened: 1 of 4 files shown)


charts/castai-cluster-controller/templates/deployment.yaml
```diff
@@ -42,6 +42,15 @@ spec:
...
       {{- end }}
       affinity:
+        podAntiAffinity:
+            requiredDuringSchedulingIgnoredDuringExecution:
+              - labelSelector:
+                  matchExpressions:
+                    - key: app.kubernetes.io/name
+                      operator: In
+                      values:
+                      - castai-cluster-controller
```
aldor007, line 52: Can you try something like this?
```suggestion
                      - {{ include "castai-agent.name" . }}
```

Source https://github.com/castai/helm-charts/blob/38c5ea89568d6789b9957865a8100516a879e880/charts/castai-cluster-controller/templates/_helpers.tpl#L47

We have some copy-paste bug in template name :D

CastAIPhil: Updated, that copy paste makes me cringe :-)

aldor007 (approved):

PR #85, opened 2022-06-01
[castai-agent] configurable rbac creation and service account name

matas-cast:


dariuspie (approved):

charts/castai-agent/values.yaml
```diff
@@ -14,6 +14,8 @@ fullnameOverride: castai-agent
 
 # Specifies whether a namespace should be created
 createNamespace: true
+# Specifies whether RBAC Clusterrole should be created
+createRBAC: true
```
saumas, line 18: to keep consistency with the world it lets name it:
```yaml
rbac:
 enabled: true
 ```

PR #89, opened 2022-06-03
[castai-spot-handler] add security context

anjmao:


charts/castai-spot-handler/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-spot-handler
 description: CAST AI spot handler daemonset chart
 type: application
-version: 0.8.0
+version: 0.9.0
```
aldor007, line 5:
```suggestion
version: 0.10.0
```

We have https://github.com/castai/helm-charts/releases/tag/castai-spot-handler-0.9.0

PR #91, opened 2022-06-03
[castai-evictor] add default security context

anjmao:


charts/castai-evictor/values.yaml
```diff
@@ -93,13 +93,10 @@ affinity:
...
       nodeSelectorTerms:
         - matchExpressions:
-            - key: "kubernetes.io/os"
-              operator: In
-              values: ["linux"]
-        - matchExpressions:
-            - key: "beta.kubernetes.io/os"
-              operator: In
-              values: ["linux"]
+            - key: kubernetes.io/os
```
viktorasm, line 96: wasn't the explicit requirement better? E.g. we would match `beta.kubernetes.io/os=windows" with this.

anjmao: The initial requirement was just to ignore windows nodes and 99% runs on linux. And `beta.kube.io/os=linux` is deprecated so we now always get warnings during components install, upgrade.

viktorasm (approved):

PR #97, opened 2022-06-06
[castai-pod-node-lifecycle] Initial commit

KarolisL:


charts/castai-pod-node-lifecycle/values.yaml
```diff
@@ -0,0 +1,148 @@
+# Values used to connect to the CAST AI API.
+api:
+  url: https://api.cast.ai
+  key:
+  clusterId:
+
+replicaCount: 3
+
+image:
+  repository: castai/pod-node-lifecycle
```
anjmao, line 10: All our components should point to google artifact registry. If we don't have it in registry yet this can be done in separate task.

charts/castai-pod-node-lifecycle/values.yaml
```diff
@@ -0,0 +1,148 @@
...
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations: {}
+
+podSecurityContext: {}
```
anjmao, line 31: Can we add default security context? (if container doesn't write any files locally) I added linter but looks that if ignores linting this chart since default values.yaml can't generate template and linter doesn't yet support custom values.

```yaml
podSecurityContext:
   readOnlyRootFilesystem: true
```

```yaml
securityContext:
  runAsNonRoot: true
  fsGroup: 1005
  runAsGroup: 1005
  runAsUser: 1005
```

KarolisL: FYI regarding linter, there's this thing: https://github.com/helm/chart-testing/blob/fbfe017b7ccbfb1fafe50d8f0962cfbba1875b8c/ct/cmd/lint.go

We can have `ci-values.yaml` which would be used during ci.

charts/castai-pod-node-lifecycle/values.yaml
KarolisL: Thanks!
We don't have it there yet, I've created a task for that.

saumas (approved):

PR #99, opened 2022-06-10
[castai-agent] use release namespace

matas-cast:


KarolisL:
Wouldn't removing the field altogether, use the release namespace?

matas-cast:
you might be right. I see that behavior in other public charts

charts/castai-agent/templates/rbac.yaml
```diff
@@ -98,15 +98,15 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: {{ include "castai-agent.serviceAccountName" . }}
-    namespace: {{ .Values.namespace }}
+    namespace: {{ .Release.Namespace }}
```
KarolisL, line 101: These are needed (related to my other comment), `metadata.namespace` are not.

matas-cast:
Tested, unfortunately when materializing chart through code it doesn't work, will have to include

KarolisL (approved):

PR #101, opened 2022-06-13
[castai-spot-handler] parameterize namespace, service account

matas-cast:


charts/castai-spot-handler/templates/namespace.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: {{ .Values.namespace }}
```
aldor007, line 5: we can remove `Values.namespace` from values, I guess? 
Do we need this namespace resource at all?

aldor007 (approved):

charts/castai-spot-handler/templates/namespace.yaml
matas-cast: yes, we can remove, thanks

Do we need this? Not sure. Helm 3 provides `--create-namespace` so in that regard we don't. But we don't pass it in any of our onboarding scripts, so might keep it

PR #105, opened 2022-06-17
[castai-agent] CPVPA uses core count

viktorasm:
To smooth out agent scaling, in case of large average node size, cpu count will be used instead to autoscale agent.


charts/castai-agent/templates/deployment.yaml
```diff
@@ -57,16 +57,15 @@ spec:
...
-          {{- else }}
+          {{- if .Values.clusterVPA.enabled }}
           {{/* we hardcode CPU req/limit for now */}}
-          resources:
             requests:
               cpu: 100m
             limits:
               cpu: 1000m
+          {{- else }}
+            {{- .Values.resources | toYaml | nindent 14 }}
```
KarolisL, line 68: nitpick: people usually use `{{- else -}}` instead of adding `{{-` to the line below.

KarolisL (approved):

PR #106, opened 2022-06-20
[castai-agent] Separate CPVPA into separate deployment

viktorasm:
In preparation to supporting autoscaler horizontal scaling, cpvpa is now a separate deployment

### DIFF
--- a/charts/castai-agent/templates/clustervpa-deployment.yaml
+++ b/charts/castai-agent/templates/clustervpa-deployment.yaml
@@ -5,16 +5,18 @@ metadata:
   name: castai-agent-cvpa
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "castai-agent.labels" . | nindent 4 }}
+    app.kubernetes.io/name: castai-agent-cvpa
+    app.kubernetes.io/instance: castai-agent-cvpa
 spec:
   replicas: 1
   selector:
     matchLabels:
-      {{- include "castai-agent.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/name: castai-agent-cvpa
   template:
     metadata:
       labels:
-        {{- include "castai-agent.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/name: castai-agent-cvpa
+        app.kubernetes.io/instance: castai-agent-cvpa
       {{- with .Values.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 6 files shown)


charts/castai-agent/values.yaml
```diff
@@ -82,3 +82,7 @@ securityContext:
   fsGroup: 1002
   runAsGroup: 1002
   runAsUser: 1002
+
+cpvpa:
```
KarolisL, line 86: Why can't we place it under existing `clusterVPA`? Having different keys might cause confusion. We could deprecate `clusterVPA` and use this instead because of better name (proportional is missing on the old one).

charts/castai-agent/templates/clustervpa-deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+      labels:
+        app.kubernetes.io/name: castai-agent-cvpa
+        app.kubernetes.io/instance: castai-agent-cvpa
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+      {{- if .Values.priorityClass.enabled }}
+      priorityClassName: {{ .Values.priorityClass.name }}
```
KarolisL, line 26: Thought: do we need priority for the autoscaler? Would it make sense to set it to a different value than agent?

charts/castai-agent/templates/clustervpa-deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+      {{- end }}
+      serviceAccountName: {{ include "castai-agent.serviceAccountName" . }}
+      automountServiceAccountToken: true
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.nodeSelector }}
```
KarolisL, line 34: same as above

charts/castai-agent/templates/clustervpa-deployment.yaml
```diff
@@ -0,0 +1,63 @@
+{{- if .Values.clusterVPA.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: castai-agent-cvpa
```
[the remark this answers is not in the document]
viktorasm: not sure what's the value of making this configurable, kinda feels more readable this way

charts/castai-agent/templates/clustervpa-deployment.yaml
viktorasm: To keep backwards-compatible behavior, didn't see any harm to align priority/node selector with agent (especially node selector). Scaling agent is vital for it's uptime, so scaler should get at least equal priority imo.

charts/castai-agent/templates/clustervpa-deployment.yaml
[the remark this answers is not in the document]
KarolisL: It is not expected to configure this field... https://github.com/castai/helm-charts/blob/51b4531d1fd8164a5f30622230636979838fec12/charts/castai-agent/templates/deployment.yaml#L4

`{{ include "castai-agent.fullname" . }}-autoscaler`

KarolisL (approved):

PR #108, opened 2022-06-29
Add apiKeySecretRef option

josefkarasek:


charts/castai-agent/values.yaml
```diff
@@ -41,14 +41,18 @@ resources: {}
...
 apiURL: "https://api.cast.ai"
 
 # Token to be used for authorizing agent access to the API
 apiKey: ""
 
+# Name of secret with Token to be used for authorizing agent access to the API
+# apiKey and apiKeySecretRef are mutually exclusive
+apiKeySecretRef: ""
+
+additionalSecretEnv: {}
```
r0kas, line 54: I would move this together with apiKey
These additional env vars would be added to same secret as apiKey

josefkarasek:
Split PRs:
* https://github.com/castai/helm-charts/pull/111
* https://github.com/castai/helm-charts/pull/112

PR #111, opened 2022-06-30
[castai-agent] add apiKeySecretRef option

josefkarasek:


charts/castai-agent/values.yaml
```diff
@@ -41,6 +41,24 @@ resources: {}
 additionalEnv:
   PPROF_PORT: "6060"
 
+additionalEnvFromRef:
```
saumas, line 44: Could we name this variable `extraEnvs` or `additionalEnvs`, something along those lines. Because it supports the usual `name: FOO value: BAR` env objects, not just `valueFrom`.

charts/castai-agent/values.yaml
```diff
@@ -41,6 +41,24 @@ resources: {}
 additionalEnv:
```
saumas, line 41: With the new variable, this `additionalEnv` is no longer needed.

charts/castai-agent/values.yaml
```diff
@@ -41,6 +41,24 @@ resources: {}
...
+  - name: AWS_ACCESS_KEY_ID
+    valueFrom:
+      secretKeyRef:
+        name: my-custom-secret
+        key: AWS_ACCESS_KEY_ID
+  - name: AWS_SECRET_ACCESS_KEY
+    valueFrom:
+      secretKeyRef:
+        name: my-custom-secret
+        key: AWS_SECRET_ACCESS_KEY
```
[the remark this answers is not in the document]
josefkarasek: I'm using `my-custom-secret` here for demonstration purposes.
The secret name is dynamically generated, hard to control in values.yaml. Any suggestion?
https://github.com/castai/helm-charts/blob/main/charts/castai-agent/templates/secret.yaml#L5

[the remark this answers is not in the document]
saumas: For demonstration you can add comments. But the `values.yaml` values must be usable. Right now the pod won't start because these secrets don't exist.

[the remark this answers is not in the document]
josefkarasek: Sorry - by _"for demonstration purposes"_ I meant for this PR review. 
I wanted to point out that the name is dynamically generated.

https://github.com/castai/helm-charts/blob/676e29fc1e418142debee289b8edef1519d79e20/charts/castai-agent/templates/secret.yaml#L5 

Now, in `additionalEnv` we set is as a constant - we're losing the link between the secret and the ref in the deployment. When the auto-generated name changes, the user will have to change the values file manually.

Do we really want this?

charts/castai-agent/templates/secret.yaml
```diff
@@ -7,7 +8,8 @@ metadata:
   labels:
     {{- include "castai-agent.labels" . | nindent 4 }}
 data:
-  API_KEY: {{ required "apiKey must be provided" .Values.apiKey | b64enc | quote }}
+  API_KEY: {{ .Values.apiKey | b64enc | quote }}
```
josefkarasek, line 11: `required` check is no longer performed. If a user does not `--set=castai.apiKey=$CASTAI_API_TOKEN`, then it's expected the user have their own secret and it is being accessed through `additionalEnv` field.

saumas: Yeah looks like we need either the secret specified directly or a secret ref. However, leave the flexible `additionalEnv` :+1:

saumas: Something like that in `values.yaml`, which could also be validated in helm helpers that either one of those exist:
```
apiKey:
  # Either value or secretKeyRef must be specified.
  #
  # value: "my hardcoded secret"
  # secretKeyRef:
  #  name: my-custom-secret
  #  key: FOOBAR
```

PR #112, opened 2022-06-30
[castai-cluster-controller] add apiKeySecretRef option

josefkarasek:

### DIFF
--- a/charts/castai-cluster-controller/templates/secret.yaml
+++ b/charts/castai-cluster-controller/templates/secret.yaml
@@ -1,3 +1,5 @@
+{{- if or .Values.apiKey .Values.castai.apiKey }}
+---
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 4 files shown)


charts/castai-cluster-controller/values.yaml
```diff
@@ -29,6 +29,10 @@ castai:
   # Token to be used for authorizing agent access to the CASTAI API
   apiKey: ""
 
+  # Name of secret with Token to be used for authorizing agent access to the API
+  # apiKey and apiKeySecretRef are mutually exclusive
```
r0kas, line 33: Add comment of required secret key `API_KEY`

charts/castai-cluster-controller/templates/secret.yaml
```diff
@@ -1,3 +1,5 @@
+{{- if or .Values.apiKey .Values.castai.apiKey }}
```
r0kas, line 1: Don't you need to nest it here as well?

`{{- if or (.Values.apiKey ) (.Values.castai.apiKey) }}`

r0kas (approved):

PR #115, opened 2022-06-30
[castai-spot-handler] add apiKeySecretRef option

josefkarasek:


charts/castai-spot-handler/values.yaml
```diff
@@ -58,3 +58,8 @@ securityContext:
   fsGroup: 1003
   runAsGroup: 1003
   runAsUser: 1003
+
+# Name of secret with Token to be used for authorizing agent access to the API
+# apiKey and apiKeySecretRef are mutually exclusive
```
aldor007, line 63: Documentation is not entirely correct (we don't have in spot-handler `apiKey`)

aldor007 (approved):

saumas (approved):

PR #122, opened 2022-07-08
[castai-agent] fix copy&paste error

josefkarasek:


charts/castai-agent/ci/secret-ref-values.yaml
```diff
@@ -0,0 +1,21 @@
...
+clusterVPA:
+  enabled: true
+resources:
+  requests:
+    cpu: 100m
+    memory: 64Mi
+  limits:
+    cpu: 1000m
+    memory: 256Mi
+isTest: true
```
josefkarasek, line 21: @saumas `isTest` is what allows the referenced secret (which is a prerequisite) to be deployed.

saumas (approved):

PR #165, opened 2022-10-25
Feature: Add commonLabels and commonAnnotations

cam3ron2:
PR to add commonLabels and commonAnnotations options to the following charts:

- castai-agent
- castai-evictor
- castai-spot-handler
- castai-cluster-controller

**Why is this feature needed:**
Certain consumers of these charts will have a need to apply labels/annotations to the resources created by these charts globally. One such use case is when using a mixture of ArgoCD/Terraform/Helm to deploy Cast. In our case, we deploy Cast via Helm/Terraform during cluster creation, after which ArgoCD is connected to the cluster to assume management of the deployed platform-level applications, of which Cast is included. 

Because ArgoCD does not *actually* use helm to deploy the helm chart when using a Helm source for an application or applicationset we are left with mismatching labels/annotations on resources that were deployed via Helm/Terraform causing the need for a force-sync (basically a deletion/re-creation) of these resources in order to reconcile.

With an option to add labels/annotations, we could instead manually set labels/annotations normally generated by Helm and smooth this process.

### DIFF
--- a/charts/castai-agent/Chart.yaml
+++ b/charts/castai-agent/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-agent
 description: CAST AI agent deployment chart.
 type: application
-version: 0.35.0
+version: 0.36.0
 appVersion: "v0.33.0"
--- a/charts/castai-agent/templates/_helpers.tpl
+++ b/charts/castai-agent/templates/_helpers.tpl
@@ -34,11 +34,23 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "castai-agent.labels" -}}
+{{- with .Values.commonLabels }}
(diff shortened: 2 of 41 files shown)


saumas:
Hey @cam3ron2, changes look good. But you need to bump the versions of all the charts.

varnastadeus:
@cam3ron2 have you tried using `podAnnotations/podLabels` which is already available for all charts?

cam3ron2:
> @cam3ron2 have you tried using `podAnnotations/podLabels` which is already available for all charts?

That option does not add labels/annotations to any objects other than pods, which is not sufficient.

cam3ron2:
> Hey @cam3ron2, changes look good. But you need to bump the versions of all the charts.

done

r0kas (approved):
Lgtm

PR #168, opened 2022-11-02
[castai-cluster-controller] cluster vpa

aldor007:
…tiAffinity

### DIFF
--- a/charts/castai-cluster-controller/templates/clustervpa-configmap.yaml
+++ b/charts/castai-cluster-controller/templates/clustervpa-configmap.yaml
@@ -1,0 +1,55 @@
+{{- if .Values.clusterVPA.enabled }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "castai-agent.fullname" . }}-cpvpa
+  labels:
+    {{- include "castai-agent.labels" . | nindent 4 }}
+  {{ if gt (len .Values.commonAnnotations) 0 -}}
+  annotations:
+    {{- include "castai-agent.annotations" . | nindent 4 }}
+  {{- end }}
+data:
+  # Increase memory requests/limits by 256Mi for every 20 nodes. round_up(nodes/nodes_per_step)*step
+  # For example, for 150 nodes: round_up(150/20)*256Mi=2048Mi
+
+  # in case of large nodes, cores-per-step will ensure that we continue to scale the agent
+  castai-cluster-contoller-autoscaler: |-
+    {
+      "cluster-controller": {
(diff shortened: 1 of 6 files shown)


charts/castai-cluster-controller/templates/clustervpa-configmap.yaml
```diff
@@ -0,0 +1,57 @@
...
+  name: {{ include "castai-agent.fullname" . }}-cluster-controller
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "castai-agent.labels" . | nindent 4 }}
+  {{ if gt (len .Values.commonAnnotations) 0 -}}
+  annotations:
+    {{- include "castai-agent.annotations" . | nindent 4 }}
+  {{- end }}
+data:
+  # Increase memory requests/limits by 256Mi for every 20 nodes. round_up(nodes/nodes_per_step)*step
```
aldor007, line 15: Copy/paste from autoscaler

charts/castai-cluster-controller/templates/clustervpa-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
+      {{- toYaml .Values.securityContext | nindent 8 }}
+      containers:
+        - name: autoscaler
+          image: k8s.gcr.io/cpa/cpvpa:{{ .Values.clusterVPA.version }}
+          command:
+            - /cpvpa
+            - --target=deployment/{{ include "castai-agent.fullname" . }}
```
aldor007, line 56: Works only for single deployment/replica

charts/castai-cluster-controller/values.yaml
```diff
@@ -54,10 +54,16 @@ additionalEnv:
...
 resources:
   requests:
-    cpu: 50m
-    memory: 100Mi
+    memory: 256Mi
+    cpu: 100m
   limits:
-    memory: 1Gi
+    memory: 256Mi
+    cpu: 100m
```
saumas, line 61: If VPA is enabled, the resources should be null because they will be adjusted automatically during runtime. Otherwise tools that depend on the diff (like argocd) will always find a mismatch.

cam3ron2:
Withdrawing, we moved away from installing castai via terraform which makes this PR of no value for us.

charts/castai-cluster-controller/values.yaml
aldor007: Done as suggested

PR #192, opened 2022-11-30
[castai-cluster-controller] Bump version v0.30.0

fals:
- Release to support aks arm64 https://github.com/castai/cluster-controller/releases/tag/v0.30.0


charts/castai-cluster-controller/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: castai-cluster-controller
 description: CAST AI cluster controller deployment chart.
 type: application
 version: 0.42.2
```
aldor007, line 5: Chart version should be bumped as well

fals: Bumped version as well

aldor007 (approved):

PR #194, opened 2022-12-05
[castai-sec-agent] update vector config

damejeras:


charts/castai-sec-agent/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: castai-sec-agent
 description: CAST AI security agent deployment chart.
 type: application
-version: 0.15.0
-appVersion: "v0.19.0"
+version: 0.15.1
```
matas-cast, line 5: I changed version 🙏

matas-cast (approved):

PR #217, opened 2023-01-13
[castai-agent] add required openshift roles and security context variables to prepare for SCCs

saumas:
Agent chart changes to make it work on OpenShift (and keep compatibility with all other providers). Allows setting custom security context for pods and containers. The default security context works on all supported providers. Also, adds required roles to get `Machine` infomation for OpenShift.


charts/castai-agent/templates/clustervpa-deployment.yaml
```diff
@@ -46,8 +46,10 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.securityContext }}
       securityContext:
-      {{- toYaml .Values.securityContext | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
```
viktorasm, line 51: what's the difference between `with .values.securityContext ....  toYaml .` and `toYaml .Values.securityContext` ?

charts/castai-agent/templates/clustervpa-deployment.yaml
```diff
@@ -61,8 +63,11 @@ spec:
           volumeMounts:
             - mountPath: /etc/config
               name: autoscaler-config
+          {{- with .Values.containerSecurityContext }}
           securityContext:
-            readOnlyRootFilesystem: true
+            {{- toYaml . | nindent 12 }}
```
viktorasm, line 68: probably also don't need a "with values.." thing

viktorasm (approved):

charts/castai-agent/templates/clustervpa-deployment.yaml
saumas: It won't even render the `securityContext` field if values are not given.

PR #240, opened 2023-03-01
[castai-agent] Agent base ram increase through cpvpa-configmap

AugustinasS:
https://castai.slack.com/archives/C049JU55L4F/p1676384691977439?thread_ts=1676299466.351299&cid=C049JU55L4F

### DIFF
--- a/charts/castai-agent/templates/clustervpa-configmap.yaml
+++ b/charts/castai-agent/templates/clustervpa-configmap.yaml
@@ -21,7 +21,7 @@ data:
       "agent": {
         "requests": {
           "memory": {
-            "base": "0",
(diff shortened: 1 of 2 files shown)


charts/castai-agent/templates/clustervpa-configmap.yaml
```diff
@@ -21,7 +21,7 @@ data:
       "agent": {
         "requests": {
           "memory": {
-            "base": "0",
```
zilvinasu, line 24: @AugustinasS  you need to bump app chart versions and do the PR title according to conventions

zilvinasu (approved):

PR #250, opened 2023-04-12
[castai-cluster-controller] Bump ctrl version v0.34.3

fals:
- Log CAST AI components deleted/evicted by controller

### DIFF
--- a/charts/castai-cluster-controller/Chart.yaml
+++ b/charts/castai-cluster-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: CAST AI cluster controller deployment chart.
 type: application
-version: 0.49.1
-appVersion: "v0.34.2"
+version: 0.49.2
+appVersion: "v0.34.3"


aldor007 (approved):

charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: CAST AI cluster controller deployment chart.
 type: application
-version: 0.49.1
-appVersion: "v0.34.2"
+version: 0.49.2
```
r0kas, line 5: Nothing has changed in the chart, why version change?

aldor007: Because we are using the 2020 flow of updating helm charts :D and we need to do it manually :D

charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
...
+appVersion: "v0.34.3"
```
r0kas, line 6: Was it a bugfix or someething new added?
If it's not a bugfix `minor` version should be changes.
As per https://semver.org/

charts/castai-cluster-controller/Chart.yaml
fals: Isn't defauolt always increase this version? I always was asked to do that

r0kas: It's not enough just to bump appVersion?

r0kas: Guess not as chart version is where release points.

charts/castai-cluster-controller/Chart.yaml
r0kas: Guess the image is released and here's not the place to talk about it. 
But please read thorugh https://semver.org/ for future reference.

r0kas (approved):

PR #276, opened 2023-06-30
[castai-agent] add DNS Policy and hostnetwork in values

CastAIPhil:


charts/castai-agent/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-agent
 description: CAST AI agent deployment chart.
 type: application
-version: 0.58.5
+version: 0.58.6
```
fals, line 5: We use SemVersion, minor is only for bug fixes

charts/castai-agent/templates/deployment.yaml
```diff
@@ -142,17 +148,6 @@ spec:
...
-            - name: {{ $k }}
-              value: "{{ $v }}"
-          {{- end }}
-          envFrom:
-            - secretRef:
-          {{- if .Values.apiKey }}
-                name: {{ include "castai-agent.fullname" . -}}
-          {{- else }}
-                name: {{ required "apiKey or apiKeySecretRef must be provided" .Values.apiKeySecretRef -}}
-          {{- end }}
```
jansyk13, line 155: Deleted by accident?

CastAIPhil: No, linter was complaining about duplicate code, this code exists at line: 81 also

CastAIPhil: You're right, it was for a different container within the pod 🤦  re-added this code.

jansyk13 (approved):

PR #278, opened 2023-06-30
[castai-evictor] add hostnetwork and dnspolicy to evictor values

CastAIPhil:


charts/castai-evictor/Chart.yaml
```diff
@@ -13,7 +13,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.21.74
+version: 0.21.75
```
fals, line 16: We use SemVersion, minor is only for bug fixes

jansyk13 (approved):

PR #283, opened 2023-07-28
[castai-evictor] Feat: adjust charts for new Eviction configuration yaml

tautvydasLiekis:


charts/castai-evictor/templates/_helpers.tpl
```diff
@@ -76,3 +76,24 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Create labels for evictor configMap
+*/}}
+{{- define "evictor.configMap.labels"}}
```
viktorasm, line 83: is this used anywhere?

tautvydasLiekis: Took a look, and nope. Fixed

saumas (approved):

PR #286, opened 2023-07-28
[castai-audit-logs-receiver] Add initial audit-logs-receiver chart

julgircast:


charts/castai-audit-logs-receiver/values.yaml
```diff
@@ -0,0 +1,125 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: us-docker.pkg.dev/castai-hub/library/audit-logs-receiver
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
+
+replicaCount: 1
```
apasyniuk, line 11: In case of persistent storage. Is it make sense to have replicas? Or it will act as hot standby? As I assume we don't have any mechanics for job sharing?

julgircast: Good point. We don't have any job sharing yet. Removed it.

apasyniuk (approved):

PR #293, opened 2023-08-04
[castai-agent] add doc

aldor007:


charts/castai-agent/values.yaml
```diff
@@ -5,10 +5,10 @@
 replicaCount: 1
 namespace: "castai-agent"
 
-# labels to add to all resources
+# commonLabels -- Labels to add to all resources.
 commonLabels: {}
 
-# annotations to add to all resources
+# commonAnnotations -- Annotations to add to all resources/
```
laimonasr, line 11: Typo / instead of .

charts/castai-agent/values.yaml
```diff
@@ -18,27 +18,28 @@ image:
 nameOverride: ""
 fullnameOverride: castai-agent
 
-# Specifies whether a namespace should be created
+# createNamespace -- Specifies whether a namespace should be created
```
laimonasr, line 21: We should probably stay consistent and either add period at the end of each value description or none of them

charts/castai-agent/values.yaml
aldor007:
```suggestion
# commonAnnotations -- Annotations to add to all resources.
```

laimonasr (approved):

PR #294, opened 2023-08-04
[castai-evictor] add doc

aldor007:


charts/castai-evictor/values.yaml
```diff
@@ -4,44 +4,44 @@
...
 commonAnnotations: {}
 
-# Specifies whether the Evictor should run in dryRun mode (Read-Only).
+# dryRyn -- Specifies whether the Evictor should run in dryRun mode (Read-Only).
 # if true, evictor will just log action(s) it would perform
 # instead of actually performing them.
 dryRun: false
 
-# Specifies whether the Evictor can behave as aggressive
+# aggressiveMode -- Specifies whether the Evictor can behave as aggressive
```
laimonasr, line 18: We should probably stay consistent and either add period at the end of each value description or none of them

PR #312, opened 2023-09-05
[castai-cluster-controller] feat: Remove cluster-controller RBAC write permissions

Narunas-K:
Breaking change: Remove cluster-controller RBAC write permissions.
In order for cluster-controller to support chartUpsert actions, cluster have to be running cluster-controller v0.38.0 version.

### DIFF
--- a/charts/castai-cluster-controller/Chart.yaml
+++ b/charts/castai-cluster-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: Cluster controller is responsible for handling certain Kubernetes actions such as draining and deleting nodes, adding labels, approving CSR requests.
 type: application
-version: 1.0.0
+version: 0.55.0
 appVersion: "v0.39.1"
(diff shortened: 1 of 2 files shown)


charts/castai-cluster-controller/templates/rbac.yaml
```diff
@@ -73,7 +73,7 @@ rules:
   - apiGroups: [ "rbac.authorization.k8s.io" ]
     resources: [ "roles", "clusterroles", "clusterrolebindings" ]
     resourceNames: [ "castai-cluster-controller", "castai-evictor", "castai-agent", "castai-spot-handler", "castai-kvisor", "castai-egressd" ]
-    verbs: [ "get", "patch", "update", "delete", "escalate" ]
```
fals, line 76: We have patch node actions, removing this would break it

fals: and I'm not sure but don't we use `update` because we need it for helm charts?

charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: Cluster controller is responsible for handling certain Kubernetes actions such as draining and deleting nodes, adding labels, approving CSR requests.
 type: application
-version: 0.54.1
+version: 1.0.0
```
aldor007, line 5: :O :)

Narunas-K: WDYT, this is a breaking change, so I think 1.0.0 makes sense ?

aldor007 (approved):

PR #355, opened 2023-11-27
[castai-cluster-controller] Add RBAC permissions to cluster-controller for castai-pod-pinner upgrades

kiriloman:


charts/castai-cluster-controller/templates/rbac.yaml
```diff
@@ -73,11 +73,11 @@ rules:
   # For installing/updating CAST AI components.
   - apiGroups: [ "rbac.authorization.k8s.io" ]
     resources: [ "roles" ]
-    resourceNames: [ "castai-cluster-controller", "castai-evictor", "castai-agent", "castai-spot-handler", "castai-kvisor", "castai-egressd" ]
+    resourceNames: [ "castai-cluster-controller", "castai-evictor", "castai-agent", "castai-spot-handler", "castai-kvisor", "castai-egressd", "castai-pod-pinner" ]
```
aldor007, line 76: you also need to extend this list https://github.com/castai/cluster-controller/blob/613c33dab377d57c1e0f4acbb42f25bfe1bcd455/helm/hook/hook.go#L36

kiriloman: Updated: https://github.com/castai/cluster-controller/pull/89

We will also need to patch the `castai-pod-pinner` webhook, so I suppose we would need to add the below, correct?

```
  - apiGroups: [ "admissionregistration.k8s.io" ] #idk
    resourceNames: [ "castai-pod-pinner" ]
    resources: [ "mutatingwebhookconfigurations" ]
    verbs: [ "patch" ]
```

charts/castai-cluster-controller/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-cluster-controller
 description: Cluster controller is responsible for handling certain Kubernetes actions such as draining and deleting nodes, adding labels, approving CSR requests.
 type: application
-version: 0.55.2
+version: 0.55.3
 appVersion: "v0.39.3"
```
aldor007, line 6: I would also update the cluster-controller version in this PR

charts/castai-cluster-controller/templates/rbac.yaml
aldor007: Yep

aldor007 (approved):

PR #356, opened 2023-11-28
[castai-cluster-controller] Allow getting castai-pod-pinner webhook using castai-cluster-controller ClusterRole

kiriloman:


aldor007 (approved):
you need to update PR title

PR #358, opened 2023-12-01
[castai-agent] scale castai-agent to 2 replicas by default

Narunas-K:


scripts/tools.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+brew install ct
+brew install yq
+brew install norwoodj/tap/helm-docs
```
tautvydasLiekis, line 5: Maybe we could use this 🤔 

`brew install --cask \
cask1 \
cask2 \
cask3 \
...
`

mjudeikis: :shrug: 
Just did oneliner. And you got other bug I missed :D name of the pacakge

mjudeikis:
@tautvydasLiekis stab seal of approval?

Issue #391, opened 2024-02-09
Pod pinner supporting HPA `autoscaling/v2`


steve-fraser:
The current API version for pod-pinner is autoscaling/v2beta1.

This was depreciated in [1.25](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v125)

kiriloman:
@steve-fraser thanks for the PR and raising this issue. However, HPA is not supported for Pod Pinner at the moment. Hence, removing it in https://github.com/castai/helm-charts/pull/396 completely.

PR #392, opened 2024-02-09
[castai-pod-pinner] fix pod pinner hpa

steve-fraser:
fixes https://github.com/castai/helm-charts/issues/391


kiriloman:
Thanks for the PR and raising this issue. However, HPA is not supported for Pod Pinner at the moment. Hence, removing it in https://github.com/castai/helm-charts/pull/396 completely.

PR #394, opened 2024-02-12
[castai-agent] remove cluster level configmap access

aweris:
This PR removes access to `configmaps` from `ClusterRole` and replaces it with configurable namespace-level RBAC.

### DIFF
--- a/charts/castai-agent/README.md
+++ b/charts/castai-agent/README.md
@@ -46,7 +46,8 @@ CAST AI agent deployment chart.
 | priorityClass.enabled | bool | `true` |  |
 | priorityClass.name | string | `"system-cluster-critical"` |  |
 | provider | string | `""` | Name of the Kubernetes service provider one of: "eks", "gks", "aks", "kops". |
-| rbac | object | `{"enabled":true}` | Specifies whether RBAC Clusterrole should be created. |
+| rbac.configmapsReadAccessNamespaces | list | `["kube-system"]` | Namespaces to be granted access to the castai-agent for configmaps read access. |
+| rbac.enabled | bool | `true` | Specifies whether a Clusterrole should be created. |
 | replicaCount | int | `2` |  |
 | resources.requests.cpu | string | `"100m"` |  |
 | resources.requests.memory | string | `"128Mi"` |  |
--- a/charts/castai-agent/templates/rbac.yaml
+++ b/charts/castai-agent/templates/rbac.yaml
@@ -289,4 +289,44 @@ subjects:
     namespace: {{ .Release.Namespace }}
 {{- end}}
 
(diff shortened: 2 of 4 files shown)


charts/castai-agent/README.md
```diff
@@ -46,7 +46,8 @@ CAST AI agent deployment chart.
 | priorityClass.enabled | bool | `true` |  |
 | priorityClass.name | string | `"system-cluster-critical"` |  |
 | provider | string | `""` | Name of the Kubernetes service provider one of: "eks", "gks", "aks", "kops". |
-| rbac | object | `{"enabled":true}` | Specifies whether RBAC Clusterrole should be created. |
+| rbac.configmapsReadAccessNamespaces | list | `["kube-system"]` | Namespaces to be granted access to the castai-agent for configmaps read access. |
+| rbac.enabled | bool | `true` |  |
```
viktorasm, line 50: missing description

aweris: had typo in `values.yaml` fixing 👍

charts/castai-agent/templates/rbac.yaml
```diff
@@ -279,4 +289,46 @@ subjects:
     namespace: {{ .Release.Namespace }}
 {{- end}}
 
+{{- if (gt (len .Values.rbac.configmapsReadAccessNamespaces) 0) }}
```
viktorasm, line 292: nitpick: is length check needed? I would expect `range` behavior would not change

```
{{range pipeline}} T1 {{end}}
	The value of the pipeline must be an array, slice, map, or channel.
	If the value of the pipeline has length zero, nothing is output;
	otherwise, dot is set to the successive elements of the array,
	slice, or map and T1 is executed. If the value is a map and the
	keys are of basic type with a defined order, the elements will be
	visited in sorted key order.
```

charts/castai-agent/values.yaml
```diff
@@ -23,9 +23,13 @@ fullnameOverride: castai-agent
 
 # createNamespace -- Specifies whether a namespace should be created.
 createNamespace: true
-# rbac -- Specifies whether RBAC Clusterrole should be created.
+
 rbac:
+  # rbac.create -- Specifies whether a Clusterrole should be created.
```
viktorasm, line 28: what is `rbac.create` ?

viktorasm (approved):

charts/castai-agent/templates/rbac.yaml
aweris: it should not, it's just a defensive reflex I guess. removing it

charts/castai-agent/values.yaml
aweris: typo, fixed it

PR #399, opened 2024-02-15
[castai-pod-pinner] add podLabels property

saumas:


charts/castai-pod-pinner/values.yaml
```diff
@@ -37,6 +37,8 @@ serviceAccount:
 
 podAnnotations: {}
 
+podLabels: {}
```
aweris, line 40: nit: since we're generating docs automatically, it would be great to have some description here.

aweris (approved):

PR #428, opened 2024-04-12
[castai-evictor] conditional volumes RBAC

jansyk13:


charts/castai-evictor/values.yaml
```diff
@@ -61,6 +61,10 @@ fullnameOverride: castai-evictor
 
 rbac:
   enabled: true
+  # When Evictor Helm chart is upgraded from Cluster Controller we don't have
+  # permissions to update ClusterRoleBinding or ClusterRole, so we need a way to skip it.
+  volumes:
+    enabled: true
```
aweris, line 67: [question]: should we enable it by default?

jansyk13: Yes, when customer installs the chart during onboarding they should get new version.

aweris (approved):
LGTM

PR #447, opened 2024-05-28
KUBE-380: Add permissions to list and delete volume attachments

stgleb:


andrejatcastai:
This should be implemented using the same pattern as we have in evictor. Otherwise we may have problems with upgrading controllers.

andrejatcastai (changes requested):
Please take a look at the way it is done for evictor.

PR #448, opened 2024-05-28
allow autoscaling via param

ValyaB:


Tsonov:
Changes make autoscaler optional so RBAC for that feature is only installed if the feature is enabled, minimizing impact.

charts/castai-cluster-controller/templates/rbac.yaml
```diff
@@ -73,7 +77,12 @@ rules:
   # For creating events.
   - apiGroups: [ "" ]
     resources: [ "events" ]
-    verbs: [ "list", "create", "patch" ]
+    verbs: [ "create", "patch" ]
+  # For CAST autoscaling CRDs management.
+  - apiGroups: [ "autoscaling.cast.ai" ]
```
aldor007, line 82: I would confirm this one with someone from WOOP

aldor007 (approved):

PR #454, opened 2024-06-07
[castai-agent] bump agent version v0.59.2

mikenorgate:


charts/castai-agent/templates/deployment.yaml
```diff
@@ -77,6 +77,18 @@ spec:
...
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: SELF_POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: SELF_POD_NODE
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
```
vladklokun, line 91: question: I understand why we need to pass namespace for this ticket, by why do we need to pass the pod and node?

charts/castai-agent/templates/deployment.yaml
```diff
@@ -77,6 +77,18 @@ spec:
             - name: shared-metadata
               mountPath: /agent-metadata
           env:
+            - name: SELF_POD_NAME
```
linkas45, line 80: why do we need these?

mikenorgate: Updated to remove as we only need namespace

linkas45 (approved):

PR #481, opened 2024-08-09
[castai-evictor] support custom root certificate authority

KarolisL:


charts/castai-evictor/templates/deployment.yaml
```diff
@@ -92,6 +92,13 @@ spec:
             - configMapRef:
                 name: {{ include "evictor.name" . }}-config
                 optional: false
+            {{- if .Values.trustedCACert }}
+            - secretRef:
+                name: {{ include "castai-agent.fullname" . -}}-tls-cert
```
mikenorgate, line 97: This should be `evictor.fullname` and you need to create the secret

charts/castai-evictor/templates/secret-tls-cert.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.trustedCACert }}
```
saumas, line 1: Where is this `trustedCACert` coming from? It should be defined in `values.yaml`.

charts/castai-evictor/templates/secret-tls-cert.yaml
```diff
@@ -0,0 +1,16 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "castai-agent.fullname" . }}-tls-cert
```
saumas, line 6: Should be `evictor.fullname`

charts/castai-evictor/templates/secret-tls-cert.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "castai-agent.labels" . | nindent 4 }}
```
saumas, line 9: `evictor.labels`

charts/castai-evictor/templates/secret-tls-cert.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{ if gt (len .Values.commonAnnotations) 0 -}}
+  annotations:
+    {{- include "castai-agent.annotations" . | nindent 4 }}
```
saumas, line 12: `evictor.annotations`

saumas (changes requested):

charts/castai-evictor/templates/secret-tls-cert.yaml
KarolisL: It is an optional value used by a small (currently, 1) subset of customers. Other services doesn't define it in `values.yaml` as well. Should I update those too, or just be consistent and don't define it anywhere?

saumas (approved):

PR #490, opened 2024-08-20
Improve helm docs generation with makefile targets

saumas:
Also ran `make docs-all` to generate missing readme's.


aldor007 (approved):

Makefile
```diff
@@ -1,11 +1,32 @@
 REPO_ROOT = $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
+.PHONY: help
+help:
+	@echo "Usage:"
+	@echo "  make lint-all                  Lint all charts in the 'charts' directory."
+	@echo "  make validate                  Validate the PR."
+	@echo "  make docs-all                  Generate helm docs for all charts in the 'charts' directory."
+	@echo "  make docs-<chart_name>         Generate helm docs for a specific chart, e.g. 'make docs-castai-agent'."
```
aweris, line 9: [suggestion]: no need to maintain manual help target. You can use something below(old example from one of my repos -- [src](https://github.com/aweris/github-actions-runner/blob/master/Makefile))


```make
.PHONY: install-tools
install-tools: ## Install tools
install-tools: vendor ; $(info $(M) installing tools)
	$(Q) $(GOINSTALL) $(shell cat tools.go | grep _ | awk -F'"' '{print $$2}')

.PHONY: help
help: ## Shows this help message
	$(Q) echo 'usage: make [target] ...'
	$(Q) echo
	$(Q) echo 'targets : '
	$(Q) echo
	$(Q) fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'| column -s: -t
	$(Q) echo
```

scripts/gen-docs.sh
```diff
@@ -0,0 +1,52 @@
...
+            exit 1
+            ;;
+    esac
+done
+
+if [[ -n "$CHART_NAME" ]]; then
+    if [[ -d "charts/$CHART_NAME" ]]; then
+        echo "Generating helm docs for chart: $CHART_NAME"
+        cd "charts/$CHART_NAME" || exit 1
+        helm-docs
```
aweris, line 35: [question] are we expecting `helm-docs` available on machine when we're running the script

aweris (approved):
LGMT 👍 

one more note, it would be amazing if we check chart docs is up to date for merge requests.

Makefile
saumas: Added dynamic help target

scripts/gen-docs.sh
saumas: Added validation that `helm-docs` exist before running the commands

PR #501, opened 2024-09-12
[castai-pod-pinner] add patch pods permission

mikenorgate:


charts/castai-pod-pinner/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: castai-pod-pinner
 description: CAST AI Pod Pinning deployment chart.
 type: application
-version: 0.7.0
-appVersion: "v0.0.16"
+version: 0.8.0
+appVersion: "v0.0.18"
```
vladklokun, line 6: question: why bump by two versions?

mikenorgate: `appVersion` is to include the changes to PP that add the patch operation
`version` because its a new chart release

vladklokun: My bad, I meant “why do we increment `appVersion` by two patch versions: from 0.0.16 to 0.0.18?

vladklokun (approved):

PR #502, opened 2024-09-12
[castai-cloud-proxy]  KUBE-503: Helm additions for cloud-proxy

Tsonov:
Note: not complete but should deploy the service in the cluster. Working on adjusting settings and secrets.

### DIFF
--- a/charts/castai-cloud-proxy/ci/test-values.yaml
+++ b/charts/castai-cloud-proxy/ci/test-values.yaml
@@ -1,0 +1,26 @@
+castai:
+  apiURL: "https://api.cast.ai"
+  grpcURL: "grpc.cast.ai:443"
+  apiKey: "gibberish"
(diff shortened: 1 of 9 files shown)


charts/castai-cloud-proxy/ci/test-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+additionalEnv:
+  LOG_LEVEL: 5
+podLabels:
+  test: test
+global:
+  commonAnnotations:
+    foo: bar
+  commonLabels:
+    faz: baz
+    x: b
```
ValyaB, line 22: should it be cleaned?

ValyaB (approved):

charts/castai-cloud-proxy/ci/test-values.yaml
Tsonov: Good catch, removed

PR #538, opened 2024-10-07
[castai-db-optimizer] update default replica count

davex98:


charts/castai-db-optimizer/templates/deployment.yaml
```diff
@@ -24,6 +24,16 @@ spec:
         app.kubernetes.io/name: {{ include "name" . }}
     spec:
       terminationGracePeriodSeconds: 60
+      affinity:
```
viktorasm, line 27: let's also add an option in helm chart values to disable host antiaffinity (enabled by default). might be needed for testing.

charts/castai-db-optimizer/README.md
```diff
@@ -0,0 +1,33 @@
...
+CAST AI database cache deployment.
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| apiKey | string | `""` | Token to be used for authorizing access to the CAST AI API.  |
+| apiURL | string | `"api.cast.ai"` | URL to the CAST AI API server. |
+| cacheGroupID | string | `""` | ID of the cache group for which cache configuration should be pulled.  |
+| proxy.logLevel | string | `"filter:info"` |  |
```
viktorasm, line 14: should we add docs for these?

viktorasm (approved):

PR #558, opened 2024-10-14
[castai-agent] add TSC and PDB to helm chart

mikenorgate:


charts/castai-agent/values.yaml
```diff
@@ -168,3 +168,13 @@ containerSecurityContext:
...
     drop:
       - ALL
+
+topologySpreadConstraints:
+  - maxSkew: 1
+    topologyKey: kubernetes.io/hostname
+    whenUnsatisfiable: DoNotSchedule
+    labelSelector:
+      matchLabels:
+        app.kubernetes.io/name: castai-agent
```
saumas, line 178: I think you should use `{{ include "castai-agent.selectorLabels" . }}` instead of the hardcoded label to support changing labels via props

saumas (changes requested):
One change requested

charts/castai-agent/values.yaml
saumas: Resolving because I didn't notice that this is in values.yaml. I think it's possible to still use variables in values.yaml but you need a tpl func which isn't worth it IMO.

saumas (approved):

PR #569, opened 2024-10-21
[castai-cluster-controller] add monitor

ValyaB:


charts/castai-cluster-controller/values.yaml
```diff
@@ -59,13 +59,16 @@ castai:
   clusterID: ""
 
 # podAnnotations -- Annotations added to each pod.
-podAnnotations: {}
+podAnnotations:
+  kubectl.kubernetes.io/default-container: "cluster-controller"
```
aldor007, line 63: I would add it directly to the deployment. This is a field for customer override

charts/castai-cluster-controller/templates/deployment.yaml
```diff
@@ -71,6 +74,9 @@ spec:
         - name: cluster-controller
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
aldor007, line 75: command is not required for this image?

charts/castai-cluster-controller/templates/deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
       dnsPolicy: {{ .Values.dnsPolicy }}
       {{- end }}
       {{- if and (.Capabilities.APIVersions.Has "scheduling.k8s.io/v1") (semverCompare ">= 1.18-0" .Capabilities.KubeVersion.Version) (.Values.priorityClass | default dict).enabled }}
+      volumes:
+        - name: shared-metadata
+          emptyDir: {}
```
aldor007, line 51: Can you please add size limit for this volume?

charts/castai-cluster-controller/templates/deployment.yaml
ValyaB: For backward compatibility, I made a controller option by default (no command specified) to cover cases when, on the latest helm version, you downscale the image version

aldor007 (approved):

stgleb (approved):
LGTM

PR #587, opened 2024-11-04
[castai-agent] Use app version v0.70.0

pijusn-cast:


pijusn-cast:
There is an unrelated issue with some Microsoft stuff:
```
    apiVersions Should Be Recent - Api versions must be the latest or under 2 years old (730 days) - API version 2022-11-01 of Microsoft.ContainerService/managedClusters is 734 days old
    apiVersions Should Be Recent - Api versions must be the latest or under 2 years old (730 days) - API version 2022-11-01 of Microsoft.KubernetesConfiguration/extensions is 734 days old
```

This PR will be blocked until it gets resolved. I'll rebase and re-run this PR then.

laimonasr (approved):

PR #602, opened 2024-11-08
[castai-audit-logs-receiver] Fix tolerations location

edaugherty:
affinity, nodeSelector, and tolerations are part of the pod spec, not the container.


charts/castai-audit-logs-receiver/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: castai-audit-logs-receiver
 description: A Helm chart for CAST AI OpenTelemetry Collector.
 type: application
 version: 0.4.2
```
davidcastai, line 5: Can you bump this to 0.4.3 so that pipelines can pass ?

davidcastai (approved):

PR #622, opened 2024-11-21
[gpu-metrics-exporter] add support for clusterIDSecretRef

spandey04:
Following the best practices of GitOps should allow minimum/no manual intervention required from the developers. When using crossplane to deploy the castai cluster resources, a secret containing the value of clusterID is stored in the cluster. These changes allow reading the value of the field "clusterID" from the secret instead of someone having to manually add it in the values files.


kisahm:
duplicate https://github.com/castai/helm-charts/pull/662

PR #628, opened 2024-11-27
[castai-db-optimizer] bump query-processor image

davex98:


charts/castai-db-optimizer/templates/_versions.tpl
```diff
@@ -1,2 +1,2 @@
 {{- define "defaultProxyVersion" -}}4.50.15{{- end -}}
-{{- define "defaultQueryProcessorVersion" -}}v0.1.10{{- end -}}
+{{- define "defaultQueryProcessorVersion" -}}v0.1.12{{- end -}}
```
benjamg, line 2: can we update the proxy version too to 4.51.0? as that is latest version of that :)

benjamg (approved):

PR #644, opened 2024-12-06
[castai-anywhere] remove redundant chart

laimonasr:


kasia-kujawa:
What is used instead of this helm chart?

kasia-kujawa (approved):

laimonasr:
A shell script will be used instead of it

PR #663, opened 2025-01-13
[castai-db-optimizer] bump QP version to 0.3.1

viktorasm:


benjamg (approved):

charts/castai-db-optimizer/values.yaml
```diff
@@ -56,6 +56,8 @@ queryProcessor:
   logLevel: "warn"
   # -- Default query-processor query cache size.
   queryCacheSize: 100000
+  # -- Number of worker threads. This should ideally be tuned around 1.5 - 2x times more than expected amount of CPU usage.
```
benjamg, line 59: not sure I agree it should ever be over 1x the number of cpus, but that also assumes we have no io work in the workers (which I think is the case)

viktorasm: due to synchronization of pre/post messages, often in a pair of workers only one of them will be progressing, hence the 2x recommendation.

PR #695, opened 2025-01-30
[castai-cluster-controller] KUBE-849: increase qps to handle spikes

furkhat:


charts/castai-agent/values.yaml
```diff
@@ -112,7 +112,11 @@ dnsPolicy: ""
 
 # Pod toleration rules.
 # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
-tolerations: {}
+tolerations:
```
krystiancastai, line 115: is this change expected?

furkhat: mistake. updated.

krystiancastai (approved):

PR #704, opened 2025-02-05
[castai-db-optimizer] support for default routing

benjamg:


r0kas (approved):
This doesn't support multi endpoint that we will do with clustered support. And potentially might be moved to config map.

PR #711, opened 2025-02-11
[castai-db-optimizer] create secret for storing API key

davex98:


charts/castai-db-optimizer/values.yaml
```diff
@@ -7,6 +7,11 @@ apiURL: "api.cast.ai"
 #
 apiKey: ""
 
+# apiKeySecretRef -- Name of secret with Token to be used for authorizing DBO access to the API
+# apiKey and apiKeySecretRef are mutually exclusive
```
benjamg, line 11: we don't forbid if both are present should we mention which takes precedent?

benjamg (approved):

charts/castai-db-optimizer/values.yaml
davex98: If both are there, then the `apiKey` will be used (it would create a secret underneath). I believe this pattern is used across all of our castware

PR #725, opened 2025-02-17
[castai-hosted-model] Add a condition for hosted model dependency

kiriloman:


charts/castai-hosted-model/Chart.yaml
```diff
@@ -2,9 +2,10 @@ apiVersion: v2
...
 description: CAST AI hosted model deployment chart. Currently, it's just a pass through to ollama chart helm.
 type: application
-version: 0.0.2
+version: 0.0.3
 appVersion: "v0.0.1"
 dependencies:
   - name: ollama
     version: 1.4.0
     repository: https://otwld.github.io/ollama-helm/
+    condition: ollama.enabled
```
Zilv1nas, line 11: Isn't this kind of a breaking change without any default values?

Zilv1nas (approved):

PR #726, opened 2025-02-17
[castai-db-optimizer] Move the proxy parameters to values.yaml and decrease evictionThreadPeriodMs to 100 ms.

kkusas:


charts/castai-db-optimizer/README.md
```diff
@@ -47,4 +53,4 @@ CAST AI database cache deployment.
 | upstreamPostgresPort | int | `5432` | Port of the upstream Postgres instance. |
 
 ----------------------------------------------
-Autogenerated from chart metadata using [helm-docs v1.11.0](https://github.com/norwoodj/helm-docs/releases/v1.11.0)
+Autogenerated from chart metadata using [helm-docs v1.14.2](https://github.com/norwoodj/helm-docs/releases/v1.14.2)
```
davex98, line 56: Did you generate it via `make docs-castai-db-optimizer`? It uses docker underneath to ensure a proper version of helm-docs.

charts/castai-db-optimizer/templates/envoy_config.yaml
```diff
@@ -20,12 +20,12 @@ data:
           path: /data/cache
           db_size_bytes: {{ .Values.resources.proxy.cacheDbSizeBytes | int }}
           max_cache_entries: {{ .Values.resources.proxy.maxCacheEntries | int }}
-          eviction_threshold: 0.5
-          stop_caching_threshold: 0.95
-          write_batch_size: 100
-          write_batch_delay_ms: 100
-          stats_thread_period_ms: 1000
-          eviction_thread_period_ms: 1000
+          eviction_threshold: {{ .Values.proxy.evictionThreshold }}}
```
davex98, line 23: Let's ensure those statements have default values equal to the old ones so the generated output won't change for existing deployments. For example: 
`{{ .Values.proxy.evictionThreshold | default 0.5 }}`

charts/castai-db-optimizer/README.md
kkusas: It said helm-docs, so this is what I used :D Not a correct version, unfortunately. I'll use docs-castai-db-optimizer from now on, thanks!

benjamg (approved):

davex98 (approved):

PR #729, opened 2025-02-18
[castai-hosted-model] Add vLLM to castai-hosted-model

kiriloman:


charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+          imagePullPolicy: IfNotPresent
+          command: ["/bin/sh", "-c", "vllm", "serve"]
+          args:
+            - {{ .Values.model | quote }}
+            - "--trust-remote-code"
+            - "--gpu-memory-utilization=1"
+            - "--task=generate"
+            - "--dtype=half"
+            - "--kv-cache-dtype=fp8"
+            - "--enable-chunked-prefill=False"
```
Zilv1nas, line 32: Is there any reason to disable this? From what I've read - enabling this provides better GPU utilization.

charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - name: "vllm"
+          image: "{{ required "Required value 'image.repository' must be defined" .Values.image.repository }}:{{ required "Required value 'image.tag' must be defined" .Values.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          command: ["/bin/sh", "-c", "vllm", "serve"]
+          args:
+            - {{ .Values.model | quote }}
+            - "--trust-remote-code"
+            - "--gpu-memory-utilization=1"
+            - "--task=generate"
+            - "--dtype=half"
```
Zilv1nas, line 30: Mostly just curious, why this and not leave it "auto"?

Zilv1nas:
How are we going to pass the HF token to this chart?

charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
kiriloman: No reason. Enabled it!

charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
kiriloman: It appears that when left with `auto`, the model may fail to run as only specific types of GPUs would be usable for it. For example, I tried running a llama3.2:3b model with `dtype=auto` and `nvidia-tesla-t4` GPU, and it would fail on startup because that GPU didn't support that specific configuration. Changing it to `half` was due to the suggestion in the vLLM logs themselves.

charts/castai-hosted-model/child-charts/vllm/templates/secret.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.hfToken }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "vllm.fullname" . }}-{{ now | date "02012006150405" }}
```
Zilv1nas, line 5: Do we need this date? I assume the `vllm.fullname` will be unique anyway.

Zilv1nas (approved):

charts/castai-hosted-model/child-charts/vllm/templates/secret.yaml
kiriloman: This date is to handle the case of same 2 models being deployed in 2 different operations. However, I will update the fullname to actually be the service name since it must be unique on our side. Just like we have for Ollama.

PR #738, opened 2025-02-24
[castai-db-optimizer] enable readiness probe by default

viktorasm:


benjamg (approved):
i dislike the need for this, feels like it shouldn't be needed, by my helm knowledge is low

PR #767, opened 2025-03-24
[castai-pod-mutator] add support for envFrom

mikenorgate:


charts/castai-pod-mutator/values.yaml
```diff
@@ -127,3 +127,6 @@ castai:
   clusterIdSecretKeyRef:
     name: ""
     key: "CLUSTER_ID"
+
+# -- Used to set additional environment variables for the pod-mutator container via configMaps or secrets.
+envFrom: []
```
michal-kopczynski, line 132: Nit: IMHO would be nice to place it next to `additionalEnv` to group somewhat related values together.

michal-kopczynski (approved):

PR #774, opened 2025-03-27
[castai-agent] handle validation scenarios when using envFrom

mikenorgate:


charts/castai-agent/templates/deployment.yaml
```diff
@@ -117,12 +117,16 @@ spec:
...
-          {{- if .Values.apiKey }}
+                {{- if .Values.apiKey }}
                 name: {{ include "castai-agent.fullname" . -}}
-          {{- else }}
+                {{- else }}
                 name: {{ required "apiKey or apiKeySecretRef must be provided" .Values.apiKeySecretRef -}}
-          {{- end -}}
+                {{- end -}}
+          {{- else if not .Values.envFrom }}
+          {{- fail "castai.apiKey or castai.apiKeySecretRef must be provided" }}
```
laimonasr, line 128: For both failure messages you might want to do the following:
1. Leave just the `apiKey` and `apiKeySecretRef` without the `castai.` prefix as I see the values are not prefixed in the other messages
2. Add `or .envFrom` to the message as it's also something that makes the conditions pass

laimonasr (approved):

PR #776, opened 2025-03-27
[castai-pod-mutator] handle validation scenarios when using envFrom

mikenorgate:


charts/castai-pod-mutator/templates/deployment.yaml
```diff
@@ -77,7 +77,7 @@ spec:
                   name: {{ required "organizationID or organizationIdSecretKeyRef must be provided" .Values.castai.organizationIdSecretKeyRef.name }}
                   key: {{ .Values.castai.organizationIdSecretKeyRef.key }}
           {{- else }}
-            {{- if and (not .Values.castai.organizationID) (not .Values.castai.configMapRef) }}
+            {{- if and (not .Values.castai.organizationID) (not .Values.castai.configMapRef) (not .Values.envFrom) }}
             {{- fail "either organizationID configMapRef or organizationIdSecretKeyRef must be provided" }}
```
michal-kopczynski, line 81: Could you also update the error messages to include that new value? There are many different configuration options, so up-to-date debugging message would be valuable.

mikenorgate: Done

michal-kopczynski (approved):

PR #777, opened 2025-03-28
[castai-cluster-controller] KUBE-1046: disable approving csr

ValyaB:


charts/castai-cluster-controller/templates/deployment.yaml
```diff
@@ -211,6 +211,10 @@ spec:
             - name: LEADER_ELECTION_LEASE_RENEW_DEADLINE
               value: "{{ .Values.leaderElectionRenewDeadline }}"
           {{- end }}
+          {{- if not .Values.leaderElectionEnabled }}
```
aldor007, line 214: Why using `leaderElectionEnabled`?

aldor007 (approved):

PR #784, opened 2025-03-31
[castai-pod-pinner] handle validation scenarios when using envFrom

mikenorgate:


charts/castai-pod-pinner/templates/secret.yaml
```diff
@@ -10,7 +10,9 @@ metadata:
     {{- include "pod-pinner.annotations" . | nindent 4 }}
   {{- end }}
 data:
+  {{- if or .Values.castai.apiKey }}
```
pijusn-cast, line 13: This `or` is probably redundant here.

```suggestion
  {{- if .Values.castai.apiKey }}
```

PR #788, opened 2025-04-03
[castai-db-optimizer] add protocol value to parser

viktorasm:


charts/castai-db-optimizer/templates/deployment.yaml
```diff
@@ -125,6 +125,8 @@ spec:
               value: {{ .Values.queryProcessor.concurrency | default "10" | quote }}
             - name: ENDPOINTS
               value: {{ .Values.endpoints | toJson | quote}}
+            - name: PROTOCOL
+              value: {{ .Values.protocol }}
```
davex98, line 129:
```suggestion
              value: {{ .Values.protocol | default "PostgreSQL" }}
```

davex98 (approved):

PR #793, opened 2025-04-08
[castai-evictor] feat(evictor): support WOOP recommendations

vladklokun:
This commit adds support for WOOP recommendations to Evictor.

It will make Evictor take WOOP recommendations into account when it runs its eviction simulations.

### DIFF
--- a/charts/castai-evictor/Chart.yaml
+++ b/charts/castai-evictor/Chart.yaml
@@ -13,7 +13,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.31.80
+version: 0.32.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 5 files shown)


charts/castai-evictor/templates/clusterrole.yaml
```diff
@@ -90,3 +90,47 @@ rules:
       - patch
       - update
   {{- end }}
+  {{- if .Values.woop.enabled }}
```
linkas45, line 93: this can go to evictor's ext helm chart and we dont even need the woop enabled/disabled

vladklokun: Moved WOOP permissions to the `castai-evictor-ext` chart without the `woop.enabled` condition, because it's more trouble to pass the `woop.enabled` value to the child Helm chart, and it has the potential to worsen UX for people updating the child chart or setting this in Terraform, where they would have to specify this parameter, too.

linkas45 (approved):

vladklokun:
Per our change of course to enable WOOP for everyone by default, I changed the default value. This PR now awaits changes in Kubecast to be merged, a bump in application version from the Evictor version bump bot and then I can rebase and merge this.

PR #794, opened 2025-04-09
[castai-cluster-controller] feat(cluster-controller): node rbac

abezard-conduit:
The goal is to be able to not grant the cluster-controller write access on nodes when only the workload autoscaler is enabled and the node autoscaler is disabled.

### DIFF
--- a/charts/castai-cluster-controller/Chart.yaml
+++ b/charts/castai-cluster-controller/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: castai-cluster-controller
 description: Cluster controller is responsible for handling certain Kubernetes actions such as draining and deleting nodes, adding labels, approving CSR requests.
 type: application
-version: 0.81.4
+version: 0.81.5
 appVersion: "v0.57.1"
 annotations:
   release-date: "2024-06-04T07:10:07"
--- a/charts/castai-cluster-controller/templates/rbac.yaml
+++ b/charts/castai-cluster-controller/templates/rbac.yaml
@@ -44,7 +44,7 @@ rules:
     resources: [ "deployments" ]
     verbs: [ "patch" ]
   {{- end}}
-  {{- if .Values.autoscaling.enabled }}
+  {{- if .Values.nodeManagement.enabled }}
   # For advertising extended resources
   - apiGroups: [ "" ]
     resources: [ "nodes/status" ]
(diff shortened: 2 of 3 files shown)


Tsonov:
Thank you for contributing to the project. 


We are reluctant to add a new flag here because it would be backwards incompatible with terraform helm installations or onboarding scripts that do not pass the flag as true and would need to be upgraded to properly install node autoscaling. 

However, there is a low hanging fruit here by extracting the following permissions outside the `autoscaling.enabled=true` check. The CRD should be owned by Cast.AI anyway so there should be no harm in always adding those permissions. This way, workload autoscaler and cluster controller can be enabled without extending permissions to node autoscaling. These are the only permissions that are required in this scenario. 

```
  # For CAST autoscaling CRDs management.
  - apiGroups: [ "autoscaling.cast.ai" ]
    resources: [ "*" ]
    verbs: [ "get", "list", "watch", "patch", "create", "delete", "update" ]
```

Feel free to adjust MR or we can open a separate one.

abezard-conduit:
@Tsonov I updated the PR with your recommendation as it achieves what I want, but I'd like to mention that it doesn't feel like the right approach to me. Permissions should ideally be clearly scoped and driven by self-explanatory config keys.

Tsonov (approved):

Tsonov:
> @Tsonov I updated the PR with your recommendation as it achieves what I want, but I'd like to mention that it doesn't feel like the right approach to me. Permissions should ideally be clearly scoped and driven by self-explanatory config keys.

I agree that some adjustment could be made and RBAC was grown organically over the years without a lot of strategy. Ideally we would have toggle for Workload autoscaling (which controls the CRD), toggle for AI optimizer, toggle for component upgrade. Added item to internal discussions but we have to make the changes backwards compatible so older versions of terraform & user scripts can still install new chart versions without breaking autoscaling; hence it is a bit larger work item.

However, this change is a fast, low-hanging fruit that shouldn't break any existing installations/workflows and in general it makes sense to have Cast's CRD management access by default. All the RBAC under the flag is related to node autoscaling (including events and CSR permissions) so it matches the flag. 

Self-upgrade will also become less of an issue as we are working on operator-based upgrade framework for Cast components outside of cluster controller. 

Again, thank you for taking this up on yourself to notice and fix!

PR #802, opened 2025-04-14
[castai-evictor] change value for security context in evictor 

vishwesh-castai:
changing  security context at container level for evictor deployment to set allowPrivilegeEscalation to false


mikenorgate:
Can you remove `podSecurityContext` from values.yaml as it's not used anymore and resolve the build errors

vishwesh-castai:
removed podSecurityContext as suggested

mikenorgate (approved):

PR #811, opened 2025-04-22
[castai-evictor] Overide envFrom.

jozuenoon:
This commit changes behavior of envFrom in Evictor helm chart.
If envFrom is defined it will completely override the default envFrom section for the evictor container.

### DIFF
--- a/charts/castai-evictor/Chart.yaml
+++ b/charts/castai-evictor/Chart.yaml
@@ -18,9 +18,9 @@ version: 0.32.7
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: 7e807b510bf3c25c735656f84d4ad25981462906
+appVersion: 4c7d149686333dff1c68c4ba860fda95c0d46a54
 annotations:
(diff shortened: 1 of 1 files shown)


charts/castai-evictor/templates/deployment.yaml
```diff
@@ -102,6 +102,10 @@ spec:
             - name: "{{ $name }}"
               value: "{{ $value }}"
             {{- end }}
+          {{- if and .Values.envFrom .Values.overrideEnvFrom}}
+          envFrom:
+            {{- toYaml .Values.envFrom | nindent 12 }}
+          {{- else }}
           envFrom:
```
laimonasr, line 109: I think we should also include `.Values.envFrom` in the original `envFrom` list if `overrideEnvFrom` is `false`

charts/castai-evictor/templates/deployment.yaml
```diff
@@ -102,6 +102,10 @@ spec:
...
             - secretRef:
```
primoz-hrovat, line 110: We could wrap these defaults with an if block. We only have one "envFrom" that is populated with user-defined values and cast defaults (if overrideEnvFrom == false).

KarolisL (approved):

PR #818, opened 2025-04-24
[castai-evictor] fix error for watch permissions

ekufta0530:


vladklokun:
Hi @ekufta0530! We cannot add new permissions to the Evictor ClusterRole directly because Cast AI upgrades Evictors we manage for our customers automatically via Cluster Controller. And when Cluster Controller upgrades Evictor to a new version that includes new permissions in Evictor's ClusterRole, the upgrades would fail because Cluster Controller does not have permission to update ClusterRoles in our customers' cluster.

As such, we cannot accept this change.

However, customers on latest versions of Evictor should not be seeing this error, since Evictor does not require this permission at runtime after `castai-evictor` Helm chart v0.31.75 and later. Are the customers running the latest version of the Evictor chart?

We also provide extended RBAC to be installed manually via the `castai-evictor-ext` [subchart](https://github.com/castai/helm-charts/blob/af8021ac306c7e6c46501f632469b683f240c481/charts/castai-evictor/child-charts/castai-evictor-ext/templates/clusterrole.yaml).

PR #820, opened 2025-04-28
NOJIRA: Change update strategy to Rolling Update

linkas45:


charts/castai-evictor/values.yaml
```diff
@@ -77,7 +77,9 @@ kubernetesClient:
 
 # updateStrategy -- Controls `deployment.spec.strategy` field.
 updateStrategy:
-  type: Recreate
+  type: RollingUpdate
+  rollingUpdate:
+    maxSurge: 0
```
pijusn-cast, line 82: You effectively still have the same behaviour in most cases (where `replicas=1`). What was being discussed in the call linked was to upscaling instance first before terminating the old pods.

pijusn-cast (changes requested):

PR #850, opened 2025-05-14
[castai-cluster-controller] Disabling if workloadManagement is disabled

steve-fraser:
A Prospect needs to be able to deploy the workload optimization in a safe way to get the data without actually giving permissions to do so.


varnastadeus:
closing in favor of https://github.com/castai/helm-charts/pull/852

PR #866, opened 2025-06-03
[castai-agent, castai-evictor] add CronJob resource permissions

ZuzanaLietavcova:


charts/castai-evictor/child-charts/castai-evictor-ext/templates/clusterrole.yaml
```diff
@@ -60,6 +60,7 @@ rules:
       - batch
     resources:
       - jobs
+      - cronjobs
```
aivarasatkocaitiscastai, line 63: Not sure if evictor needs this 🤔

ZuzanaLietavcova: I don't know either, just assuming if it needs jobs it may need also cronjobs. I will ask about it.

JolantaSk (approved):

UndeadRat22 (approved):

PR #909, opened 2025-07-04
[castai-db-optimizer] switch for new style of routing

viktorasm:


benjamg (approved):
I would have inverted the logic to be useOldRouting but this is fine too

viktorasm:
> I would have inverted the logic to be useOldRouting but this is fine too

this is a bit hard for existing deployments (upgrading would require adding a value, and custom defaults are tricky in helm)

PR #912, opened 2025-07-08
[castai-db-optimizer] provide mechanism for injecting custom TLS certificates

mbezhanov:


charts/castai-db-optimizer/templates/envoy_config.yaml
```diff
@@ -92,9 +92,9 @@ data:
                     common_tls_context:
                       tls_certificates:
                         - certificate_chain:
-                            filename: "cert.pem"
+                            filename: "{{ include "proxy.tls.certname" . }}"
```
viktorasm, line 95: can we make a single helper for "tls_certificates" section?

charts/castai-db-optimizer/values.yaml
```diff
@@ -83,6 +83,8 @@ proxy:
   drainTimeSeconds: 60
   # -- Extra network debug logging.
   networkDebug: false
+  # -- Whether to use a Kubernetes TLS Secret (true) or the default self-signed certificate (false).
+  useCustomTlsCertificate: false
```
viktorasm, line 87: maybe instead of "true/false" let's just make it `tlsCertificateSecret` , and mention in the comment it needs to have "tls.crt" and "tls.key" values.

viktorasm (approved):

PR #916, opened 2025-07-14
[castai-watchdog] refactor chart and add support for HTTP notification webhooks

Trojan295:
- add support for notifications via HTTP webhooks
- refactor the config management a bit, use a config file instead of environment variables, where it was possible

### DIFF
--- a/charts/castai-watchdog/Chart.yaml
+++ b/charts/castai-watchdog/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-watchdog
 description: CAST AI Watchdog
 type: application
-version: 0.1.0
-appVersion: "v0.0.1"
+version: 0.2.0
+appVersion: "8ea6cee6bfdb98894bb7190eec1f7ea8ceab5e3c"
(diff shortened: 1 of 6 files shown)


charts/castai-watchdog/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-watchdog
 description: CAST AI Watchdog
 type: application
-version: 0.1.0
-appVersion: "v0.0.1"
+version: 0.2.0
+appVersion: "8ea6cee6bfdb98894bb7190eec1f7ea8ceab5e3c"
```
cirisked, line 6: Can u explain why it's not a semver?

Trojan295: I initially wanted to use SemVer, but then I noticed that evictor is using also commit hashes and our pipelines are automatically building and publishing images from [kubecast's master](https://gitlab.com/castai/kubecast/-/blob/master/services/watchdog/.gitlab-ci.yml?ref_type=heads#L20).

We could manually tag the kubecast with a SemVer tag like `watchdog/vx.y.z` and get SemVer images, but didn't see a reason, not sure if customers care about it.

cirisked (approved):

PR #923, opened 2025-07-16
[castai-db-optimizer] fix inconsistent endpoint ports POLY-745 v2

benjamin-rood:


charts/castai-db-optimizer/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: castai-db-optimizer
 description: CAST AI database cache deployment.
 type: application
-version: 0.37.0
+version: 0.37.1-rc.1
```
benjamg, line 5: I thought we gave up on rc builds?

charts/castai-db-optimizer/templates/service.yaml
```diff
@@ -17,8 +17,8 @@ spec:
     {{- range $index, $endpoint := .Values.endpoints}}
     {{- if empty $endpoint.name }}
     - name: endpoint-{{$index}}
-      port: {{$endpoint.port}}
-      targetPort: {{$endpoint.port}}
+      port: {{$endpoint.servicePort}}
+      targetPort: {{$endpoint.targetPort}}
```
benjamg, line 21: won't these need the reverted changed re-reverted first?

benjamg (approved):

charts/castai-db-optimizer/templates/service.yaml
benjamin-rood: Oh, you mean on the polyscale side? Yes.

charts/castai-db-optimizer/Chart.yaml
benjamin-rood: I'm gunshy after yesterday. I don't want this to risk getting into the wild

PR #927, opened 2025-07-17
[gpu-metrics-exporter] Role/Cluster Role, CastAI API URL configuration

roman-castai:


roman-castai:
Need to make a change to the underlying repo, closing it here

PR #951, opened 2025-08-13
[castai-db-optimizer] fix broken service ports in service chart

benjamin-rood:
Fix service port mapping in service.yaml template to use correct endpoint.port instead of endpoint.servicePort. This ensures Kubernetes services expose the correct envoy internal ports and properly route to upstream database ports.

- Change service port from {{$endpoint.servicePort}} to {{$endpoint.port}}
- Add clarifying comments for port vs targetPort usage
- Bump chart version to 0.42.1 (production release)

The fix resolves service port mismatches where endpoints showed correct values but generated services had invalid port configurations.


benjamg:
https://github.com/castai/helm-charts/pull/949/files please combine so we don't confuse things

PR #956, opened 2025-08-14
[castai-hosted-model] Add model downloader

Zilv1nas:
- Use RunAI streamer to load model weights directly into GPU.
- Add model downloader init container to download LoRa adapters from GCS
- Add S3 proxy sidecar to make GCS compatible with RunAI streamer, as it currently works with S3 instead of GCS.

I will move this chart to `ai-optimizer-proxy` later, as reviewing the chart changes otherwise would be difficult.

### DIFF
--- a/charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
+++ b/charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
@@ -22,20 +22,20 @@ spec:
         {{- end }}
     spec:
       volumes:
-        {{ if .Values.tensorParallelSize }}
+        {{- if .Values.tensorParallelSize }}
         # vLLM needs to access the host's shared memory for tensor parallel inference.
         - name: shm
           emptyDir:
             medium: Memory
             sizeLimit: "10Gi"
-        {{ end }}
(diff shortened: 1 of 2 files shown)


charts/castai-hosted-model/child-charts/vllm/templates/deployment.yaml
```diff
@@ -130,6 +148,56 @@ spec:
...
+              mountPath: /models
+      initContainers:
+{{- if .Values.loraAdapter.name }}
+        - name: model-downloader
+          image: "{{ .Values.modelDownloader.image.repository }}:{{ .Values.modelDownloader.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          args:
+            - download
+          env:
+            - name: STORAGE_TYPE
```
magaman384, line 161: nit: Passing the storage related env vars and volume mounts is basically the same as in S3 proxy. For now the code duplicity is not a big deal, but once we will start adding other providers this will bloat. Would be a good idea to refactor it a bit, so we can reuse that part.

Zilv1nas: Updated.

PR #959, opened 2025-08-15
[castai-db-optimizer] bump `dbo-proxy` image to latest version

mbezhanov:


charts/castai-db-optimizer/templates/_versions.tpl
```diff
@@ -1,2 +1,2 @@
-{{- define "defaultProxyVersion" -}}v4.63.0{{- end -}}
```
benjamg, line 1: update to v4.64.14

benjamg (approved):

benjamg (changes requested):
change version sorry to say this in two comments, couldn't work out how to change approve -> request change

PR #982, opened 2025-09-02
[castai-evictor] Update live migrations rbac.

jozuenoon:
1. Update live migration rbac in evictor.

### DIFF
--- a/charts/castai-evictor/child-charts/castai-evictor-ext/templates/clusterrole.yaml
+++ b/charts/castai-evictor/child-charts/castai-evictor-ext/templates/clusterrole.yaml
@@ -75,12 +75,6 @@ rules:
   # ------------------------------------------------
   # Live Migration: retrieving and creating migrations
   # ------------------------------------------------
-  - apiGroups:
-      - cast.ai
-    resources:
-      - migrations
-    verbs:
-      - get
-      - list
-      - watch
-      - create
+  - apiGroups: [ "live.cast.ai" ]
+    resources: [ "*" ]
+    verbs: [ "get", "list", "watch", "patch", "create", "delete", "update" ]
--- a/charts/castai-evictor/values.yaml
+++ b/charts/castai-evictor/values.yaml
@@ -47,7 +47,7 @@ liveMigration:
   # to another node without interrupting these workloads.
   # Evictor will only consider LIVE migrations for workloads that explicitly request it.
   # as long as they can be scheduled somewhere else.
-  enabled: false
+  enabled: true
   # If true, Evictor will use a cache for its LIVE-specific client.
   # This should reduce the amount of read requests issued to the API server when working with LIVE migrations.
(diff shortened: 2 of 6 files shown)


charts/castai-evictor/values.yaml
```diff
@@ -47,7 +47,7 @@ liveMigration:
   # to another node without interrupting these workloads.
   # Evictor will only consider LIVE migrations for workloads that explicitly request it.
   # as long as they can be scheduled somewhere else.
-  enabled: false
+  enabled: true
```
vladklokun: chore: this flag makes Evictor check for Live Migration capabilities unconditionally

While the check first makes sure we have appropriate permissions, I can see that the new permissions in this PR are broader than previously and they're more than what we check in Evictor.

I suggest you test the following:
* Set up a clean cluster with Live Migrations, install Evictor as usual — check if Evictor works with CLM.
* Remove some permissions from the `castai-evictor` RBAC, to simulate customers that have older RBAC clusters that don't have the `castai-evictor-ext` chart updated — check if Evictor gracefully reports errors and does the usual evictions.
* Remove the `castai-evictor` RBAC altogether, so we have no permissions to access migrations to simulate customers that have old Evictors that don't wish to have extended capabilities — check if Evictor gracefully reports errors and does the usual evictions.
* Remove the Live Migration components, especially the CRD itself and restart Evictor — check if Evictor gracefully reports errors and does the usual evictions. We had an incident with WOOP where not having the CRD in the cluster resulted in an incident. Evictor should correctly detect that we don't have the CRD, meaning no permission to create it, report it and use the usual evictions.

I’ll send you the relevant Evictor code through a different channel so you can understand this better and expand on the tests.

One scenario we currently don't cover is if Live CRDs are removed from the cluster AND Evictor is not restarted. Then, Evictor will still think it is capable of Live Migrations, but with the CRDs gone that is no longer true.

charts/castai-evictor/child-charts/castai-evictor-ext/templates/clusterrole.yaml
```diff
@@ -75,12 +75,6 @@ rules:
...
-      - migrations
-    verbs:
-      - get
-      - list
-      - watch
-      - create
\ No newline at end of file
+  - apiGroups: [ "live.cast.ai" ]
+    resources: [ "*" ]
+    verbs: [ "get", "list", "watch", "patch", "create", "delete", "update" ]
```
vladklokun, line 80: question: why do we need broader permissions that previously?

Is this intended to just catch all possible permissions for the future? Or is there a legitimate need for them now?

In any case, please make sure that Evictor's current check for permissions is up-to-date. I’ll send the relevant code through a different channel.

vladklokun (approved):
Permissions are fine, so as we figured out, all that's left is to bump the `-ext` chart version.

PR #994, opened 2025-09-09
[castai-db-optimizer] fix pooling secrets logic

benjamg:


charts/castai-db-optimizer/templates/pgcat-config.yaml
```diff
@@ -2,18 +2,35 @@
 {{- if not .Values.pooling.databases }}
 {{- fail "pooling.databases cannot be empty when pooling is enabled. Please provide at least one database name." }}
 {{- end }}
+{{- $secretName := "" }}
+{{- if .Values.pooling.credentialsSecretRef }}
+{{- $secretName = .Values.pooling.credentialsSecretRef.name }}
```
mbezhanov, line 7: The `name` seems redundant.

```suggestion
{{- $secretName = .Values.pooling.credentialsSecretRef}}
```

charts/castai-db-optimizer/templates/pgcat-config.yaml
```diff
@@ -2,18 +2,35 @@
...
+{{- else }}
+{{- $secretName = printf "%s-pooling-credentials" (include "name" .) }}
+{{- end }}
+{{- $credentials := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
+{{- $username := "" }}
+{{- $password := "" }}
+{{- if $credentials }}
+{{- $username = ($credentials.data.username | b64dec) }}
+{{- $password = ($credentials.data.password | b64dec) }}
```
mbezhanov, line 16: Because the secret stores these as `USERNAME` and `PASSWORD` (uppercase) for this to work correctly it has to be:

```suggestion
{{- $username = ($credentials.data.USERNAME | b64dec) }}
{{- $password = ($credentials.data.PASSWORD | b64dec) }}
```

Otherwise, you'd get:

> Error: UPGRADE FAILED: template: castai-db-optimizer/templates/pgcat-config.yaml:15:46: executing "castai-db-optimizer/templates/pgcat-config.yaml" at <b64dec>: invalid value; expected string

charts/castai-db-optimizer/templates/secret-creds.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.pooling .Values.pooling.enabled }}
+{{- if and .Values.pooling .Values.pooling.enabled (not .Values.pooling.credentialsSecretRef) }}
```
mbezhanov, line 1: As the values from this `Secret` are only referenced in [charts/castai-db-optimizer/templates/pgcat-config.yaml](charts/castai-db-optimizer/templates/pgcat-config.yaml) (which itself creates a `Secret`), this `Secret` seems redundant. You can instead use these values directly in the other `Secret` as follows:
```
{{- $username := .Values.pooling.username }}
{{- $password := .Values.pooling.password }}
```

So this:

```
{{- if not .Values.pooling.databases }}
{{- fail "pooling.databases cannot be empty when pooling is enabled. Please provide at least one database name." }}
{{- end }}
{{- $secretName := "" }}
{{- if .Values.pooling.credentialsSecretRef }}
{{- $secretName = .Values.pooling.credentialsSecretRef }}
{{- else }}
{{- $secretName = printf "%s-pooling-credentials" (include "name" .) }}
{{- end }}
{{- $credentials := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
{{- $username := "" }}
{{- $password := "" }}
{{- if $credentials }}
{{- $username = ($credentials.data.username | b64dec) }}
{{- $password = ($credentials.data.password | b64dec) }}
{{- else }}
{{- $username = .Values.pooling.username }}
{{- $password = .Values.pooling.password }}
{{- end }}
```

Can become:

```
{{- if not .Values.pooling.databases }}
{{- fail "pooling.databases cannot be empty when pooling is enabled. Please provide at least one database name." }}
{{- end }}
{{- $username := "" }}
{{- $password := "" }}
{{- if .Values.pooling.credentialsSecretRef }}
{{- $secretName := .Values.pooling.credentialsSecretRef }}
{{- $credentials := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
{{- $username = ($credentials.data.username | b64dec) }}
{{- $password = ($credentials.data.password | b64dec) }}
{{- else }}
{{- $username = .Values.pooling.username }}
{{- $password = .Values.pooling.password }}
{{- end }}
```

charts/castai-db-optimizer/templates/pgcat-config.yaml
```diff
@@ -2,18 +2,35 @@
...
+kind: Secret
 metadata:
   name: {{ include "name" $ }}-pgcat-config-{{ $endpointName }}
   labels:
     {{- include "labels" $ | nindent 4 }}
   annotations:
     {{- include "annotations" $ | nindent 4 }}
-data:
+type: Opaque
+stringData:
```
mbezhanov, line 33: This line defaults to generating a `server_lifetime` in scientific notation:

```
server_lifetime = {{ $.Values.pooling.serverLifetime | default 86400000 }}
```

Because of that, `pgcat` won't start due to the following error:

```
2025-09-10T09:11:27.085880Z  INFO ThreadId(01) pgcat: Welcome to PgCat! Meow. (Version 1.2.0)    
2025-09-10T09:11:27.089056Z ERROR ThreadId(01) pgcat::config: Could not parse config file: TOML parse error at line 8, column 19
  |
8 | server_lifetime = 8.64e+07
  |                   ^^^^^^^^
invalid type: floating point `86400000`, expected u64
    
2025-09-10T09:11:27.089120Z ERROR ThreadId(01) pgcat: Config parse error: BadConfig   
```

You can change as follows to ensure the value isn't converted to scientific notation:

```
server_lifetime = {{ $.Values.pooling.serverLifetime | default "86400000" | int }}
```

charts/castai-db-optimizer/templates/secret-creds.yaml
mbezhanov: Okay, in that case please make sure to change `USERNAME` to `username` and `PASSWORD` to `password` otherwise the `Secret` won't work properly with the logic implemented in `charts/castai-db-optimizer/templates/pgcat-config.yaml` as it expects both of these to be lowercase.

PR #995, opened 2025-09-10
[castai-hosted-model] Add vLLM CI tests

marcinrybinski-castai:


charts/castai-hosted-model/child-charts/vllm/values.yaml
```diff
@@ -57,39 +57,67 @@ dtype: "half"
 kvCacheDtype: "auto" # Same as default in the vLLM engine. It will read models config and use the same dtype for cache.
 maxLoraRank: 128
 
+# -- Startup probe configuration
 startupProbe:
+  # -- Enable or disable startup probe
+  enabled: true
```
marcinrybinski-castai, line 63: Had to introduce `enabled` params to be able to disable probing in CI tests.

Zilv1nas: What was wrong with probes enabled in ci tests?

charts/castai-hosted-model/ci/test-values.yaml
```diff
@@ -1,19 +1,44 @@
...
-
   service:
+    name: vllm-test
     port: 11434
     type: ClusterIP
-  ingress:
-    enabled: false
+  image:
+    repository: "magglish/vllm"
+    tag: "cpu-no-optim"
```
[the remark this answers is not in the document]
marcinrybinski-castai: After:
1. Disabling probes
2. Configuring vLLM in a way that it won't download the model and run the service
the custom image is no longer needed and we can work on an image provided by vLLM.

PR #1049, opened 2025-10-21
[castai-agent] Only create service account if Values.serviceAccount.create is true

sionescu:
Currently, a service account is created if `Values.rbac.enabled` is true but that's incorrect: it should be created if `Values.serviceAccount.create` is true.

### DIFF
--- a/charts/castai-agent/templates/rbac.yaml
+++ b/charts/castai-agent/templates/rbac.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.enabled }}
+{{- if .Values.serviceAccount.create }}
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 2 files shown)


primoz-hrovat:
@sionescu Thanks for the contribution! Our workflows are failing, because Chart version needs to be bumped and PR title must start with `[castai-agent] `. Would you mind to update the PR please?

sionescu:
@primoz-hrovat Done.

charts/castai-agent/templates/rbac.yaml
```diff
@@ -418,4 +420,4 @@ subjects:
 
 {{- end }}
 
-{{- end }}
+{{- end }} // if .Values.rbac.enabled
```
primoz-hrovat: This `// ` is not a valid comment. Remove or use something similar to `{{{/* this is a comment directive in Helm */}}` or regular YAML comments `#`.

primoz-hrovat:
@sionescu I overlooked the small problem with the use of comments. Please update

charts/castai-agent/templates/rbac.yaml
sionescu: Done.

primoz-hrovat (approved):

PR #1075, opened 2025-11-06
[castai-evictor] fix rendering with wrong KarpenterMode ref

mszostok:
```
█ Installing castai-evictor.
Release "castai-evictor" does not exist. Installing it now.
Error: template: castai-evictor/templates/deployment.yaml:76:31: executing "castai-evictor/templates/deployment.yaml" at <.Values.KarpenterMode.enabled>: nil pointer evaluating interface {}.enabled

```


aldor007 (approved):
Thx!

PR #1091, opened 2025-11-13
[castai-spot-handler] KUBE-1574: Don't try to deref the parent object when rendering clusterID

Tsonov:
Makes it more obvious that the whole struct must be present, not a specific value from it.


Tsonov:
Closing, we lose the benefit of "documented fields in values.yaml" and it's a minor change to reset `.name` instead.

PR #1105, opened 2025-11-26
WIRE-1872 - add option to set chart version from release tag

oskarwojciski:
## Summary

  - Add `chart-version-from-tag` option to `remote-release` action
  - When enabled, chart version is set from release tag (e.g., `v0.1.0` → `version: 0.1.0`)
  - Default behavior unchanged (auto-increment patch) for backwards compatibility

  ## Problem

  Currently the `remote-release` action always auto-increments the patch version:
  - `0.0.22` → `0.0.23` → `0.0.24`

  This makes it impossible to release minor or major versions without manually editing `Chart.yaml` beforehand.

  ## Solution

  Add optional `chart-version-from-tag` input:

  ```yaml
  - uses: castai/helm-charts/.github/actions/remote-release@main
    with:
      chart-version-from-tag: "true"  # version from tag

  | Mode    | Input | Behavior                                 |
  |---------|-------|------------------------------------------|
  | Default | false | Auto-increment patch (existing behavior) |
  | New     | true  | Use release tag as version               |

  Backwards Compatibility

  ✅ Fully backwards compatible - default is false, existing workflows unchanged.

  Testing

  # Default mode (auto-increment)
  python bump_chart.py Chart.yaml v0.1.0
  # 0.0.22 → 0.0.23

  # New mode (from tag)
  python chart_version_from_tag.py Chart.yaml v0.1.0
  # 0.0.22 → 0.1.0

### DIFF
--- a/.github/actions/remote-release/chart_version_from_tag.py
+++ b/.github/actions/remote-release/chart_version_from_tag.py
@@ -1,12 +1,26 @@
 import re
 import sys
 
+# SemVer regex pattern
+SEMVER_PATTERN = re.compile(
+    r"^(\d+)\.(\d+)\.(\d+)"
+    r"(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?"
+    r"(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?$"
+)
 
-def update_version(chart_yaml, new_version, chart_yaml_path):
+
+def is_valid_semver(version: str) -> bool:
+    """Check if the version string follows SemVer format."""
+    # Strip optional 'v' prefix before validation
+    version_to_check = version.lstrip("v")
+    return bool(SEMVER_PATTERN.match(version_to_check))
+
+
+def update_version(chart_yaml: str, new_version: str, chart_yaml_path: str) -> str:
     """Update the version field in Chart.yaml content."""
     match = re.search(r"^version:\s*(.+)$", chart_yaml, re.IGNORECASE | re.MULTILINE)
     if not match:
-        raise Exception(f"No 'version' field found in {chart_yaml_path}")
+        raise ValueError(f"No 'version' field found in {chart_yaml_path}")
 
     current_version = match.group(1)
     print(f"Current version: {current_version}")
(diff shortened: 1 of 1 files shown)


.github/actions/remote-release/chart_version_from_tag.py
```diff
@@ -0,0 +1,48 @@
...
+new_version = release_tag.lstrip("v")
+
+with open(chart_yaml_path, "r") as chart_file:
+    chart_yaml = chart_file.read()
+
+updated_yaml = chart_yaml
+
+# Update version to match release tag
+match = re.search(r"version:\s*(.+)", chart_yaml)
+if match:
```
niccolocast: What happens if version match is None? I see that it's a silent failure, should we throw an exception and fail properly?

oskarwojciski: Hmm, yeah it might be good to add a safe check, but AFAIK Chart.yaml without a version is invalid and will fail in packaging

oskarwojciski: i will make the regexp case insensitive in case someone will put `Version` in chart

.github/actions/remote-release/chart_version_from_tag.py
radekska: suggestion: the script could be structured with functions to make the script more readable with `main()` function under `if __name__ == "__main__":`

.github/actions/remote-release/chart_version_from_tag.py
```diff
@@ -0,0 +1,75 @@
+import re
+import sys
+
+
+def update_version(chart_yaml, new_version, chart_yaml_path):
```
radekska: nit: [type](https://docs.python.org/3/library/typing.html) annotations could be added

.github/actions/remote-release/chart_version_from_tag.py
```diff
@@ -0,0 +1,75 @@
...
+        raise Exception("Chart.yaml path should be passed as first argument")
+
+    if len(sys.argv) < 3 or sys.argv[2] == "":
+        raise Exception("Release tag should be passed as second argument")
+
+    chart_yaml_path = sys.argv[1]
+    release_tag = sys.argv[2]
+
+    # Extract version from tag (v0.1.2 -> 0.1.2)
+    new_version = release_tag.lstrip("v")
```
radekska: Should we validate tag against SemVer structure to disallow setting random strings as versions?

oskarwojciski: Done, I have added a simple regex check because I didn't want to rely on external libraries (like https://github.com/python-semver/python-semver) in GitHub Actions.

PR #1108, opened 2025-11-26
[castai-db-optimizer] drop experimental connection pooling

davex98:


benjamg (approved):
all the pgcat work :(

PR #1111, opened 2025-11-27
[castai-hosted-model] Add reasoning parser and increase timeouts for Kimi-K2-Thinking

marcinrybinski-castai:
Add reasoning parser and increase timeouts for Kimi K2 Thinking deployment.

### DIFF
--- a/charts/castai-hosted-model/Chart.lock
+++ b/charts/castai-hosted-model/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: vllm
   repository: file://child-charts/vllm
-  version: 0.0.29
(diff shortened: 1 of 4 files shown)


charts/castai-hosted-model/child-charts/vllm/values.yaml
```diff
@@ -64,8 +64,8 @@ startupProbe:
   enabled: true
   initialDelaySeconds: 20
   periodSeconds: 6
-  # Allow for up to 20 minutes of startup time
-  failureThreshold: 200
+  # Allow for up to 60 minutes of startup time
+  failureThreshold: 600
```
krzysztofreczek: Wow, this is pretty long. Is this really needed?

PR #1310, opened 2026-03-17
[castai-evictor] Add evictorconfig CRD.

jozuenoon:


charts/castai-evictor/templates/_helpers.tpl
```diff
@@ -111,10 +111,24 @@ Resolve imagePullSecrets: merge global.imagePullSecrets with local imagePullSecr
...
 {{ .Values.customConfig | nindent 4 }}
+  {{- else if kindIs "slice" .Values.customConfig }}
+    {{- /* Structured array mode */ -}}
+{{ "evictionConfig:" | nindent 4 }}
+{{- toYaml .Values.customConfig | nindent 6 }}
+  {{- else if .Values.customConfig.evictionRules }}
+    {{- /* Map mode with nested evictionRules */ -}}
+{{ "evictionConfig:" | nindent 4 }}
+{{- toYaml .Values.customConfig.evictionRules | nindent 6 }}
+  {{- end }}
```
laimonasr, line 132: nit: difficult to read due to condition entries changing indents, but rendering entries falling back to the start of the line

laimonasr (approved):

charts/castai-evictor/templates/_helpers.tpl
jozuenoon: I was trying to split this visually somehow. Not sure if flattening would help with reading.

PR #1363, opened 2026-04-13
adding global tolerations

steve-fraser:
adding global tolerations


oanatmaria:
Hi, @steve-fraser you can close this PR as I raised separate PRs for each component as we have different teams handling different components and it will be easier to review/merge and release. I also handled kvisor and live as they are in separate repos

PR #1366, opened 2026-04-14
[castai-cluster-controller] add global toleration support

oanatmaria:


charts/castai-cluster-controller/templates/_helpers.tpl
```diff
@@ -83,6 +83,17 @@ Create the name of the service account to use
...
+Resolve tolerations: prefer .Values.tolerations (if non-empty), fall back to .Values.global.tolerations.
+*/}}
+{{- define "cluster-controller.tolerations" -}}
+{{- $global := .Values.global | default dict -}}
+{{- $tolerations := .Values.tolerations | default (dig "tolerations" list $global) -}}
+{{- if $tolerations -}}
+{{- toYaml $tolerations -}}
+{{- end -}}
+{{- end }}
+
```
Trojan295, line 95: Can you use this, looks a bit simpler?
```suggestion
{{- define "cluster-controller.tolerations" -}}
{{- with (.Values.tolerations | default .Values.global.tolerations) }}
{{ toYaml . }}
{{- end }}
{{- end }}

```

Trojan295 (approved):

PR #1371, opened 2026-04-14
[castai-umbrella] remove castai-evictor from kent subchart

saumas:
KENT mode includes the Continuous Rebalancer (CRB) which subsumes evictor's functionality, making evictor redundant in that profile. Removed `castai-evictor` as a dependency from the `kent `subchart and updated the umbrella README template accordingly. All other modes (autoscaler, autoscaler-anywhere) are unaffected.

### DIFF
--- a/charts/castai-umbrella/charts/kent/Chart.yaml
+++ b/charts/castai-umbrella/charts/kent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kent
 description: Wrapper chart for CAST AI Kent profile.
 type: application
-version: 0.12.0
+version: 0.11.0
 dependencies:
   - name: castai-agent
     version: "0.151.0"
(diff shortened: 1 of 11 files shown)


charts/castai-umbrella/charts/kent/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kent
 description: Wrapper chart for CAST AI Kent profile.
 type: application
-version: 0.10.1
+version: 0.12.0
```
mindaugasCast: 0.10 to 0.12 why not to 0.11? 😆 
Approving anyway, just interested why

mindaugasCast (approved):

charts/castai-umbrella/charts/kent/Chart.yaml
saumas: Accident :)

PR #1372, opened 2026-04-14
[castai-workload-autoscaler] add global toleration support

oanatmaria:


charts/castai-workload-autoscaler/Chart.yaml
```diff
@@ -3,5 +3,5 @@ name: castai-workload-autoscaler
 description: CAST AI workload autoscaler.
 type: application
 
-version: 0.1.204
+version: 0.1.205
```
UOndro, line 6: This is automatically updated on release of the new version

UOndro (dismissed):

PR #1374, opened 2026-04-14
[castai-workload-autoscaler-exporter] add global toleration support

oanatmaria:


charts/castai-workload-autoscaler-exporter/Chart.yaml
```diff
@@ -3,5 +3,5 @@ name: castai-workload-autoscaler-exporter
 description: CAST AI workload autoscaler exporter.
 type: application
 
-version: 0.0.125
```
UOndro, line 6: Same here -> This is automatically updated when releasing the new version

UOndro (dismissed):

PR #1395, opened 2026-04-28
[castai-db-optimizer] generate self-signed cert for ProxySQL to support MySQL StarTLS relay

davex98:
---
### What changed
Enabled TLS encryption for ProxySQL frontend connections by generating self-signed certificates at pod startup and configuring ProxySQL to terminate TLS. Also ensures Envoy consistently uses TLS transport for upstream connections regardless of ProxySQL configuration.

### Why
Fixes missing TLS certificate setup for ProxySQL, allowing database clients to establish encrypted connections to the cache layer.

### Key changes
- `templates/deployment.yaml`: Added `openssl` command in the init container to generate 2048-bit RSA self-signed certificates (`proxysql.crt` and `proxysql.key`) valid for 10 years
- `templates/proxysql-config.yaml`: Added `have_ssl=true`, `ssl_cert`, and `ssl_key` configuration parameters to enable SSL listening in ProxySQL
- `templates/envoy_config.yaml`: Removed conditional `if not` block that previously excluded TLS transport socket configuration when ProxySQL was enabled
- `Chart.yaml`: Bumped chart version from `0.76.0` to `0.77.0`

### Impact
Clients connecting to ProxySQL must now use TLS. The generated certificates are self-signed; clients may need to configure SSL mode to allow self-signed certificates or skip verification. This may be a breaking change for clients previously connecting over plaintext.

### DIFF
--- a/charts/castai-db-optimizer/Chart.yaml
+++ b/charts/castai-db-optimizer/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: castai-db-optimizer
 description: CAST AI database cache deployment.
 type: application
-version: 0.76.0
+version: 0.77.0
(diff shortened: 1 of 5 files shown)


westwoodra2 (approved):
LGTM

PR #1409, opened 2026-05-08
[castai-db-optimizer] support RDS IAM Authentication

davex98:
---
### What changed
Added support for AWS RDS IAM authentication and replaced the generic `useIAMAuth` toggle with provider-specific flags for GCP Cloud SQL and AWS RDS.

### Why
The previous `useIAMAuth` option only covered GCP Cloud SQL. Introducing separate, provider-specific flags enables the DB agent to support AWS RDS IAM authentication and makes the configuration intent explicit per cloud provider.

### Key changes
- `values.yaml`: Replaced `database.useIAMAuth` with `database.useCloudSQLIAMAuth` and added `database.useRDSIAMAuth`.
- `templates/deployment.yaml`: Replaced the `DATABASE_USE_IAM_AUTH` environment variable with `DATABASE_USE_CLOUD_SQL_IAM_AUTH` and `DATABASE_USE_RDS_IAM_AUTH`.
- `templates/db-user-secret.yaml`: Added validation ensuring `useCloudSQLIAMAuth` and `useRDSIAMAuth` are mutually exclusive, and that `database.password` is unset when either IAM option is enabled. Updated the password requirement check.
- `Chart.yaml` & `README.md`: Bumped chart version to `0.18.0`.
- `templates/_versions.tpl`: Bumped default DB agent version to `v0.18.1`.

### Impact
**Breaking change**: Existing deployments using `database.useIAMAuth` for GCP Cloud SQL IAM must update their configuration to use `database.useCloudSQLIAMAuth` before upgrading. The two new provider-specific flags are mutually exclusive and cannot be enabled at the same time.

### DIFF
--- a/charts/castai-db-agent/values.yaml
+++ b/charts/castai-db-agent/values.yaml
@@ -14,12 +14,14 @@ database:
   port: 5432
   # -- Username for database authentication. For IAM auth, use the format: "service-account-name@project-id.iam"
   username: ""
-  # -- Password for database authentication (not required when useIAMAuth is true)
+  # -- Password for database authentication (not required when useCloudSQLIAMAuth is true)
   password: ""
-  # -- Name of secret with database credentials (username and password). Not required when using useIAMAuth with inline username.
+  # -- Name of secret with database credentials (username and password). Not required when using useCloudSQLIAMAuth or useRDSIAMAuth with inline username.
   credentialsSecretRef: ""
   # -- Enable IAM authentication for database connection (GCP Cloud SQL IAM). Requires Workload Identity setup. See IAM_AUTH_SETUP.md for details.
-  useIAMAuth: false
+  useCloudSQLIAMAuth: false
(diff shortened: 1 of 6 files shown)


charts/castai-db-agent/values.yaml
```diff
@@ -14,12 +14,14 @@ database:
   port: 5432
   # -- Username for database authentication. For IAM auth, use the format: "service-account-name@project-id.iam"
   username: ""
-  # -- Password for database authentication (not required when useIAMAuth is true)
+  # -- Password for database authentication (not required when useCloudSQLIAMAuth is true)
```
westwoodra2: Comment on line 19 says "Not required when using useCloudSQLIAMAuth or useRDSIAMAuth" and this comment only mentions useCloudSQLIAMAuth.

westwoodra2 (approved):

PR #1435, opened 2026-05-22
[castai-chart-upgrader] Add chart upgrader chart

vbaltrusaitis-castai:
This PR adds a chart upgrader running on a cron schedule. The updater is generic and can target arbitrary charts and releases. But it's intended for inclusion in kent umblrella chart. Adding it to kent umbrella chart will come in a separate PR, when the upgrader is released.

By default, the updater get cluster admin access. One can override that by providing a custom service account.

---
### What changed
Introduces the `castai-chart-upgrader` Helm chart, which deploys a CronJob to automatically upgrade a target Helm release to the latest chart version on a configurable schedule. Also adds an end-to-end test that validates upgrading from a previous umbrella chart version, and updates the E2E workflow trigger to run on PRs affecting this new chart.

### Why
Provides an in-cluster mechanism to keep Helm releases current without relying on external CI pipelines or manual `helm upgrade` commands.

### Key changes
- `.github/workflows/e2e-umbrella.yaml`: Expands the `pull_request` trigger condition to also match PR titles containing `castai-chart-upgrader`.
- `charts/castai-chart-upgrader/`: New Helm chart for automated, scheduled Helm upgrades.
- `charts/castai-chart-upgrader/templates/cronjob.yaml`: CronJob that runs `helm upgrade` with `--reset-then-reuse-values`, supporting both standard Helm repositories and OCI registries with optional authentication via `existingSecret` or inline `username`/`password`.
- `charts/castai-chart-upgrader/templates/rbac.yaml`: Provisions a `ServiceAccount`, wildcard `ClusterRole`, and `ClusterRoleBinding` so the upgrader can manage arbitrary resources declared by target charts.
- `charts/castai-chart-upgrader/values.yaml`: Exposes configuration for `schedule`, `targetRelease`, `chart` repository/auth, `upgrade` behavior (`atomic`, `wait`, `timeout`), and job resources.
- `test/e2e/Makefile`: Adds `e2e-chart-upgrader` target.
- `test/e2e/chart_upgrader_test.go`: E2E test that installs an older `castai` umbrella release, deploys the upgrader targeting it, manually triggers the CronJob, and asserts the release upgrades to a newer version successfully.

### Impact
- The upgrader is granted cluster-admin-equivalent permissions by default. Operators requiring restricted access must set `.Values.rbac.create` and `.Values.serviceAccount.create` to `false` and supply a custom `ServiceAccount`.
- E2E tests will now execute for PR titles containing either `castai-umbrella` or `castai-chart-upgrader`.


vbaltrusaitis-castai:
kimchi review

charts/castai-chart-upgrader/templates/rbac.yaml
```diff
@@ -0,0 +1,42 @@
...
+# and provide any other service account.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "chart-upgrader.serviceAccountName" . }}
+  labels:
+    {{- include "chart-upgrader.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["*"]
+    resources: ["*"]
```
aldor007, line 25: Could we limit it to castware namespaces?

aldor007 (approved):

charts/castai-chart-upgrader/templates/rbac.yaml
vbaltrusaitis-castai: @aldor007 The issue is that then the upgrade will fail if the new chart version requires permissions outside the namespace. And it's not an edge-case: the agent may want to read a new kind of CR, pod mutator gets a feature that required deleting pods / restarting workloads, etc.

We could allow various different limitations that would prevent certain classes of upgrades. But maybe for now, it's better to 1) allow everything, 2) leave the escape hatch by allowing to use custom SA, and 3) see what feedback / requests we get on this.

Mindaugas suggested that this upgrader shouldn't be enabled by default in umbrella, but maybe explicitly enabled with a flag in the command generated in the UI.

PR #1438, opened 2026-05-22
[castai-evictor] add serviceAccount.rbac toggle for cluster-scoped RBAC

CastAIPhil:
Add a `serviceAccount.rbac` boolean toggle (default: `true`) that conditionally creates ClusterRole and ClusterRoleBinding resources in both the parent castai-evictor chart and the castai-evictor-ext child chart.

When set to `false`, SecOps teams can manage cluster-scoped RBAC independently while the chart continues to render namespace-scoped Role/RoleBinding resources.

**Files changed:**
- `values.yaml`: add `serviceAccount.rbac` field
- `templates/clusterrole.yaml`: guard with `serviceAccount.rbac`
- `templates/clusterrolebinding.yaml`: guard with `serviceAccount.rbac`
- `Chart.yaml`: update dependency condition + bump chart version

---
### What changed
Adds a `global.rbac.clusterScoped.enabled` toggle (defaulting to `true`) to the Evictor Helm chart and its `castai-evictor-ext` dependency. When set to `false`, the chart skips creating ClusterRole and ClusterRoleBinding resources.

### Why
Supports deployments in clusters where cluster-scoped RBAC is restricted, externally managed, or replaced by namespace-scoped permissions.

### Key changes
- `values.yaml` (main chart and child chart): Introduced `global.rbac.clusterScoped.enabled` (defaults to `true`).
- `templates/clusterrole.yaml` and `templates/clusterrolebinding.yaml`: Wrapped resource creation in a conditional `dig "rbac" "clusterScoped" "enabled"` check.
- `child-charts/castai-evictor-ext/templates/clusterrole.yaml` and `clusterrolebinding.yaml`: Applied the same conditional guards.
- `README.md` (both charts): Documented the new `global.rbac.clusterScoped.enabled` parameter.
- `Chart.yaml`: Bumped chart version to `0.35.70` and updated release annotations.

### Impact
Default behavior is unchanged. If `global.rbac.clusterScoped.enabled` is set to `false`, equivalent permissions must be provided through other means or Evictor will lack required cluster access.

### DIFF
--- a/charts/castai-evictor/README.md
+++ b/charts/castai-evictor/README.md
@@ -37,9 +37,10 @@ Cluster utilization defragmentation tool
 | extraVolumeMounts | list | `[]` | Used to set additional volume mounts. |
 | extraVolumes | list | `[]` | Used to set additional volumes. |
 | fullnameOverride | string | `"castai-evictor"` |  |
-| global | object | `{"castai":{"apiKeySecretRef":""},"imagePullSecrets":[],"registry":""}` | Global values propagated from parent charts. |
+| global | object | `{"castai":{"apiKeySecretRef":""},"imagePullSecrets":[],"rbac":{"clusterScoped":{"enabled":true}},"registry":""}` | Global values propagated from parent charts. |
 | global.castai.apiKeySecretRef | string | `""` | Name of a pre-existing Secret containing the CAST AI API key. Takes effect when apiKeySecretRef is not set locally. |
 | global.imagePullSecrets | list | `[]` | Image pull secrets applied to all pods. Merged with local imagePullSecrets. |
+| global.rbac.clusterScoped.enabled | bool | `true` | Enable cluster-scoped RBAC resources (ClusterRole, ClusterRoleBinding). Set to false to disable cluster-scoped RBAC if using namespace-scoped permissions. |
 | global.registry | string | `""` | Container registry prefix for all images (e.g. "my-registry.example.com"). When set, it is prepended to all image repositories. |
 | hostNetwork.enabled | bool | `false` | Enable host networking. |
 | ignorePodDisruptionBudgets | bool | `false` | Specifies whether the Evictor should ignore Pod Disruption Budgets (PDBs). If true, evictor will attempt to evict pods even if it would violate a PDB. Use with caution as this may disrupt application availability guarantees. |
--- a/charts/castai-evictor/values.yaml
+++ b/charts/castai-evictor/values.yaml
@@ -21,6 +21,11 @@ global:
     apiKeySecretRef: ""
   # global.imagePullSecrets -- Image pull secrets applied to all pods. Merged with local imagePullSecrets.
   imagePullSecrets: []
+  # global.rbac.clusterScoped.enabled -- Enable cluster-scoped RBAC resources (ClusterRole, ClusterRoleBinding).
+  # Set to false to disable cluster-scoped RBAC if using namespace-scoped permissions.
+  rbac:
+    clusterScoped:
+      enabled: true
 
 # dryRyn -- Specifies whether the Evictor should run in dryRun mode (Read-Only).
 # if true, evictor will just log action(s) it would perform
--- a/charts/castai-evictor/Chart.lock
+++ b/charts/castai-evictor/Chart.lock
@@ -3,4 +3,4 @@ dependencies:
   repository: file://child-charts/castai-evictor-ext
   version: 0.5.0
 digest: sha256:6b104203eb2f32cea8879803a479d4df6e91bf4b014a25208566b253eb6bdb80
-generated: "2026-03-09T20:53:30.464703+01:00"
+generated: "2026-06-04T12:44:22.675839-04:00"
--- a/charts/castai-evictor/Chart.yaml
+++ b/charts/castai-evictor/Chart.yaml
@@ -14,21 +14,19 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
-version: 0.35.68
+version: 0.35.69
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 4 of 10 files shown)


CastAIPhil:
@kimchi-review[bot] Thanks for the review. Re: the dependency condition semantics — this is intentional.

-  is **not defined** in the default , so Helm falls through to .
- Default behavior:  → child chart is enabled.
- With  → child chart is suppressed.
- If an operator explicitly sets , that takes precedence, which is the desired escape hatch.

Also fixed:
- Amended commit to include updated  (resolves lint-test failure)
- Updated PR title to comply with  prefix (resolves validate failure)

charts/castai-evictor/README.md
```diff
@@ -79,6 +79,7 @@ Cluster utilization defragmentation tool
 | serviceAccount.annotations | object | `{}` | Annotations to add to the service account. |
 | serviceAccount.create | bool | `true` | Specifies whether a service account should be created. |
 | serviceAccount.name | string | `""` | The name of the service account to use. If not set and create is true, a name is generated using the fullname template. |
+| serviceAccount.rbac | bool | `true` | Specifies whether ClusterRole and ClusterRoleBinding should be created. Set to false if you want to manage cluster-scoped RBAC independently. |
```
oleksandrsan: Nit: I think it might be better to make RBAC block independent from serviceAccount:
```
rbac:
  create: true
  scope: cluster | namespace # can be added in future to allow namespaced RBAC instead

serviceAccount:
  create: true
  name: ""
  ...
```

and then:
```
{{- if .Values.rbac.create }}
# ClusterRole + ClusterRoleBinding
{{- end }}
```

Nesting rbac under serviceAccount implies a parent-child relationship but in the current implementation setting `serviceAccount.create=false` does not disable RBAC rule creation.

charts/castai-evictor/values.yaml
```diff
@@ -136,6 +136,9 @@ rbac:
 serviceAccount:
   # serviceAccount.create -- Specifies whether a service account should be created.
   create: true
+  # serviceAccount.rbac -- Specifies whether ClusterRole and ClusterRoleBinding should be created.
+  # Set to false if you want to manage cluster-scoped RBAC independently.
+  rbac: true
```
laimonasr: I think this could be scoped and structured a bit better to fit the layout of the parent and child chart

Currently this is a parent chart's value that also controls whether the child chart is installed while we also have a dedicated child chart toggle `castai-evictor-ext.enabled`. Value `rbac` is also a boolean which sounds like it should disable the full rbac control, but it only affects cluster scoped resources

I would suggest to use `global` vars that control cluster scoped rbac since global values are available in both parent and child charts they should be a better fit without coupling parent chart's values with child chart

```
global:
  rbac:
    clusterScopeRbac:
      enabled: true/false
```

then use this value to toggle cluster scoped rbac resources. It's also extensible as it is no longer an umbrella boolean

charts/castai-evictor/README.md
oleksandrsan: E.g. ingress-nginx follows this approach: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L877

CastAIPhil: It was done that way to stay consistent with charts like cluster-controller which has serviceaccount creation and clusterrole tied together: https://github.com/castai/helm-charts/blob/main/charts/castai-cluster-controller/templates/rbac.yaml#L1

charts/castai-evictor/values.yaml
CastAIPhil: This would again be a major difference between how we have handled this scenario in other charts. 

Do we have coding standards that dictate how to handle these situations? Both of the comments in this PR are directly at odds with how cluster-controller does the exact same toggle.

oleksandrsan: There is already a difference between how cluster-controller and woop charts handle role creation:
1. cluster-controller only creates cluster role + binding when `serviceAccount.create`==true without the ability to disable only RBAC
2. woop charts create cluster role + binding when `serviceAccount.rbac`==true

It would be great to have everything follow the same approach, but the issue is that one approach can't solve all problems, especially the one that was selected initially without taking into account other chart requirements.

The evictor chart has a child/extension chart that other examples don't have.
In our internal discussion we agreed that it is important to have the chart condition unchanged `condition: castai-evictor-ext.enabled` and not rely on specific parameters to allow adding non-RBAC resources to it in the future. Enabling/disabling concrete resources should be done through variables, not by disabling the whole child chart.

Passing values to the child chart without using [global](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values) values requires specifying them two times during chart instantiation: once for the main chart (`serviceAccount.rbac=false`) and once for child chart (`castai-evictor-ext.rbac=false`)

Instead, it is might easier to use a global block for setting RBAC parameters once for all charts in the tree.

oleksandrsan: Proposed approach:

Move the RBAC toggle to `global.rbac.clusterScoped.enabled` (default true) and use dig for safe access in templates: `{{- if (dig "rbac" "clusterScoped" "enabled" true (.Values.global | default dict)) }}`.

This:
- lets a single value control cluster-scoped RBAC across both parent and child charts without coupling the ext chart's lifecycle to the toggle
- keeps the condition: castai-evictor-ext.enabled unchanged
- follows the existing global.* pattern used for apiKey/registry/tolerations in the umbrella.

The same convention can be adopted incrementally by other charts (spot-handler, cluster-controller, workload-autoscaler, etc.) to give umbrella operators a single knob for cluster-scoped RBAC.

charts/castai-evictor/README.md
primoz-hrovat: Check the latest comment thread

charts/castai-evictor/values.yaml
CastAIPhil: @oleksandrsan Can you give it a check now?

oleksandrsan (approved):

PR #1441, opened 2026-05-25
fix: remove orphaned castai-agent CNAB config and rename CNAB job

saumas:
The `castai-agent` chart was moved to another repository, but its CNAB config directory was left behind, causing the Release CNAB workflow to fail on every `castai-agent-*` tag because it tries to read a Chart.yaml file that no longer exists. This failed CNAB check is named `release`, which collides with the Release Charts workflow check and breaks umbrella chart updates for unrelated charts that share the same commit SHA. This PR removes the orphaned config and renames the CNAB job to `release-cnab` so the umbrella updater only waits on the correct checks.


aldor007 (approved):

.github/workflows/release-cnab.yaml
laimonasr, line 1: Do we still need this workflow?

PR #1455, opened 2026-06-08
[castai-chart-upgrader] Preserve RBAC for upgrader job

oanatmaria:
---
### What changed
Adds the `helm.sh/resource-policy: keep` annotation to the RBAC resources in the `castai-chart-upgrader` Helm chart so they are retained during Helm uninstalls and upgrades. Bumps the chart version to `0.1.1`.

### Why
Prevents the updater job's `ServiceAccount`, `ClusterRole`, and `ClusterRoleBinding` from being deleted, ensuring required permissions remain available if the job executes during a chart uninstall or upgrade.

### Key changes
- `charts/castai-chart-upgrader/templates/rbac.yaml`: Added `helm.sh/resource-policy: keep` annotation to the `ServiceAccount`, `ClusterRole`, and `ClusterRoleBinding`.
- `charts/castai-chart-upgrader/Chart.yaml`: Bumped chart version from `0.1.0` to `0.1.1`.
- `charts/castai-chart-upgrader/README.md`: Updated version badge to `0.1.1`.

### Impact
RBAC resources will now persist in the cluster after `helm uninstall`; manual cleanup is required if complete removal is desired.

### DIFF
--- a/charts/castai-chart-upgrader/Chart.yaml
+++ b/charts/castai-chart-upgrader/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-chart-upgrader
 description: Auto-updates charts via CronJob
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: "0.1.0"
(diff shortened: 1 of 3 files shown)


oskarwojciski (approved):

zamai (approved):
Uh so this would fix the issue for keeping the airbag of the updater in place.
But I think we need to look closer into the design.

PR #1463, opened 2026-06-10
[castai-chart-upgrader] Limit chart-upgrader to upgrading to chart of the same name

vbaltrusaitis-castai:
If the release you're trying to upgrade has a different chart name in the metadata than the one you're upgrading to, the upgrade job will fail with an error message.

---
### What changed
Adds a pre-upgrade safeguard that verifies the target Helm release exists and matches the expected chart name before running `helm upgrade`. Also makes `targetRelease` and `targetNamespace` explicit, required configuration values with new defaults.

### Why
Prevents the CronJob from accidentally modifying the wrong release or failing silently when the target is missing or was installed with a different chart.

### Key changes
- `templates/cronjob.yaml`: Added `helm get metadata` verification step that validates the target release exists in `targetNamespace` and its chart name matches `.Values.chart.name` before invoking `helm upgrade`; changed the upgrade namespace from `.Release.Namespace` to the new `chart-upgrader.targetNamespace` helper.
- `templates/_helpers.tpl`: `chart-upgrader.targetRelease` is now required and no longer falls back to `.Release.Name`; added new required `chart-upgrader.targetNamespace` helper.
- `values.yaml`: Added `targetNamespace` with default `"castai-agent"` and changed `targetRelease` default from `""` to `"castai"`.
- `ci/test-values.yaml`: Added explicit `targetRelease` and `targetNamespace` values for CI testing.
- `Chart.yaml`: Bumped chart version to `0.1.2`.
- `README.md`: Updated version badge and parameter documentation.

### Impact
**Breaking change**: The upgrader no longer defaults to the installing release name and namespace. `targetRelease` defaults to `"castai"` instead of falling back to `.Release.Name`, and `targetNamespace` defaults to `"castai-agent"` instead of using `.Release.Namespace`. Update these values before upgrading to ensure the correct release is targeted.

### DIFF
--- a/charts/castai-chart-upgrader/Chart.yaml
+++ b/charts/castai-chart-upgrader/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: castai-chart-upgrader
 description: Auto-updates charts via CronJob
 type: application
-version: 0.1.1
+version: 0.1.2
 appVersion: "0.1.0"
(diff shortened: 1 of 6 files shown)


vbaltrusaitis-castai:
Closing in favor of https://github.com/castai/helm-charts/pull/1464

PR #1464, opened 2026-06-10
[castai-chart-upgrader] guard subchart installs

zamai:
## Summary
- add a default-on target release guard to castai-chart-upgrader before running helm upgrade
- fail safely when the target release is owned by a wrapper chart or when the target chart is installed as a dependency
- bump castai-chart-upgrader to 0.2.0 and wire it through Kent / umbrella
- document that the updater is only for standalone CAST AI umbrella installs

## Validation
- ct lint --config ct.yaml --charts charts/castai-chart-upgrader
- ct lint --config ct.yaml --charts charts/castai-umbrella
- git diff --check
- local kind: wrapped chart with non-castai top-level chart fails before upgrade and keeps deployment/castai-agent
- local kind: wrapped chart named castai with dependency castai fails before upgrade and keeps deployment/castai-agent
- local kind: direct public umbrella install upgrades successfully

## Context
This prevents the CSU-5400 failure mode where the updater runs helm upgrade castai castai/castai against a release installed from a customer wrapper chart, reuses wrapper-shaped values, and deletes CAST AI resources.

---
### What changed
Adds pre-upgrade validation to the `castai-chart-upgrader` CronJob to block upgrades when the target chart is not the top-level release or is listed as a dependency. Also updates documentation and bumps the chart version to `0.2.0`.

### Why
Upgrading a chart that is installed as a subchart can corrupt the parent release state and is not supported by standard Helm dependency management. This guard ensures the upgrader only acts on standalone installations.

### Key changes
- `templates/cronjob.yaml`: Added validation script before `helm upgrade`:
  - Fetches target release metadata with `helm get metadata`.
  - Verifies the release's top-level chart equals `.Values.chart.name`.
  - Uses `jq` to detect if the target chart appears in the release's `dependencies`.
  - Fails with a descriptive error if the chart is a dependency or the chart name mismatches.
- `values.yaml`: Updated comments for `targetRelease` and `chart.name` to warn against subchart use and recommend tools like Renovate, Argo CD Image Updater, or Flux Image Automation.
- `Chart.yaml`: Bumped version from `0.1.1` to `0.2.0`.
- `README.md`: Synchronized version badge and parameter descriptions with `values.yaml`.

### Impact
- Existing CronJob instances that target a subchart or misconfigured release will now exit with an error instead of attempting an upgrade.
- Users managing CAST AI as a dependency must switch to parent-chart automation before adopting this version.

### DIFF
--- a/charts/castai-chart-upgrader/README.md
+++ b/charts/castai-chart-upgrader/README.md
@@ -1,6 +1,6 @@
 # castai-chart-upgrader
 
-![Version: 0.1.1](https://img.shields.io/badge/Version-0.1.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.1.0](https://img.shields.io/badge/AppVersion-0.1.0-informational?style=flat-square)
+![Version: 0.2.0](https://img.shields.io/badge/Version-0.2.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.1.0](https://img.shields.io/badge/AppVersion-0.1.0-informational?style=flat-square)
 
 Auto-updates charts via CronJob
 
@@ -12,7 +12,7 @@ Auto-updates charts via CronJob
 | chart.auth.existingSecret | string | `""` |  |
 | chart.auth.password | string | `""` |  |
 | chart.auth.username | string | `""` |  |
-| chart.name | string | `"castai"` |  |
+| chart.name | string | `"castai"` | Chart name to upgrade (e.g., castai, kent). This must match the target release's top-level chart name. Use this chart only when the target chart is installed as a standalone Helm release. If CAST AI is installed as a dependency of another chart, use dependency automation such as Renovate, Argo CD Image Updater, or Flux Image Automation instead. |
 | chart.oci | bool | `false` |  |
 | chart.repository | string | `"https://castai.github.io/helm-charts"` |  |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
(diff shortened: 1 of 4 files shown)


charts/castai-chart-upgrader/templates/cronjob.yaml
```diff
@@ -102,8 +102,39 @@ spec:
...
+              TARGET_RELEASE="{{ include "chart-upgrader.targetRelease" . }}"
+              TARGET_CHART="{{ .Values.chart.name }}"
+
+              {{- if .Values.upgrade.verifyTargetRelease }}
+              echo "Validating target release '$TARGET_RELEASE' before upgrade"
+              RELEASE_METADATA="$(helm get metadata "$TARGET_RELEASE" \
+                --namespace "{{ .Release.Namespace }}")"
+              CURRENT_CHART="$(printf '%s\n' "$RELEASE_METADATA" | awk -F': ' '/^CHART:/ {print $2; exit}')"
+              CURRENT_DEPENDENCIES="$(printf '%s\n' "$RELEASE_METADATA" | awk '/^DEPENDENCIES:/ {sub(/^DEPENDENCIES:[[:space:]]*/, ""); print; exit}')"
+
```
vbaltrusaitis-castai, line 113: k8s image has `jq` installed: you could do `--output json` and use `jq` to extract this data instead of bespoke parsing with `awk`.

charts/castai-chart-upgrader/README.md
```diff
@@ -31,6 +31,7 @@ Auto-updates charts via CronJob
 | tolerations | list | `[]` |  |
 | upgrade.atomic | bool | `true` |  |
 | upgrade.timeout | string | `"10m"` |  |
+| upgrade.verifyTargetRelease | bool | `true` |  |
```
vbaltrusaitis-castai: I'm not sure this flag is even needed, given the limited goals of the upgrader. But I won't object.

vbaltrusaitis-castai (approved):

charts/castai-chart-upgrader/README.md
zamai: yes, droped, LLM slopy :D

PR #1472, opened 2026-06-18
[castai-hosted-model] Introduce sglang

igor-susic1:
---
### What changed
Adds optional SGLang inference engine support to the `castai-hosted-model` Helm chart as an alternative to vLLM. The two engines are mutually exclusive per release.

### Why
To allow deploying hosted models using SGLang (`sglang.launch_server`) in addition to the existing vLLM backend.

### Key changes
- `Chart.yaml` and `Chart.lock`: Added `sglang` sub-chart dependency (version `0.0.1`) gated by `sglang.enabled`; bumped parent chart version to `0.0.55`.
- `child-charts/sglang/`: New sub-chart providing `Deployment`, `Service`, `Secret`, and helper templates.
  - `values.yaml` exposes `model.name`, `model.hfToken`, `tensorParallelSize`, `shm.sizeLimit`, `persistence.hostPath`, `extraArgs`, and configurable readiness/liveness/startup probes.
  - `deployment.yaml` runs `sglang.launch_server` with `/dev/shm` mounted when tensor parallelism is enabled and supports model weight caching via `hostPath` or `emptyDir`.
- `values.yaml`: Added `sglang.enabled` (default `false`) and `podAutoscaler.port` (default `8000`).
- `templates/pod-autoscaler.yaml`: Replaced hardcoded metrics port with `.Values.podAutoscaler.port`.
- `templates/validate.yaml`: Added validation to fail templating if both `vllm.enabled` and `sglang.enabled` are `true`.

### Impact
Existing vLLM-only releases are unaffected because SGLang is disabled by default. Attempting to enable both engines simultaneously now fails at template time with a clear error instead of producing conflicting resources.


igor-susic1:
@getkimchi review

charts/castai-hosted-model/child-charts/sglang/templates/_helpers.tpl
```diff
@@ -0,0 +1,41 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "sglang.fullname" -}}
```
pmateusz, line 6: The SGLang helper uses `service.name` as the resource name, but the new `sglang` values do not define `service.name`. With the minimal documented enablement path, Helm renders blank names and selector labels for the SGLang `Service` and `Deployment`, which Kubernetes will reject.

Repro:

```bash
helm dependency build charts/castai-hosted-model

helm template test charts/castai-hosted-model \
  --set vllm.enabled=false \
  --set sglang.enabled=true \
  --set sglang.model.name=meta-llama/Llama-3.1-8B-Instruct
```

The rendered manifests contain empty values like:

```yaml
metadata:
  name:
labels:
  model.aibrix.ai/name:
  app.kubernetes.io/name:
```

I’d suggest either adding a non-empty `sglang.service.name` default, making it explicitly `required`, or switching the helper to a standard Helm fullname derived from release/chart name.

charts/castai-hosted-model/values.yaml
```diff
@@ -1,6 +1,9 @@
 vllm:
   # -- Specifies if vLLM model should be deployed
   enabled: true
+sglang:
+  # -- Specifies if SGLang model should be deployed. Mutually exclusive with vllm.enabled.
```
pmateusz, line 5: The values say `sglang.enabled` is mutually exclusive with `vllm.enabled`, but the chart leaves them as independent dependency conditions, and `vllm.enabled` defaults to `true`. That means setting only `sglang.enabled=true` still leaves vLLM enabled, and a fully configured install can render both backends at once.

Repro:

```bash
helm dependency build charts/castai-hosted-model

helm template test charts/castai-hosted-model \
  --set sglang.enabled=true \
  --set sglang.service.name=my-sglang \
  --set sglang.model.name=meta-llama/Llama-3.1-8B-Instruct \
  --set vllm.service.name=my-vllm \
  --set vllm.model.name=meta-llama/Llama-3.1-8B-Instruct
```

I’d suggest adding Helm validation in the parent chart so this fails early with a clear message instead of silently deploying both engines:

```yaml
{{- if and .Values.vllm.enabled .Values.sglang.enabled -}}
{{- fail "Only one model engine can be enabled: set either vllm.enabled=true or sglang.enabled=true, not both." -}}
{{- end -}}
```

That keeps the existing values shape. Given `vllm.enabled` currently defaults to `true`, the docs/examples should also show SGLang users setting `vllm.enabled=false`.

pmateusz (changes requested):

charts/castai-hosted-model/child-charts/sglang/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - name: shm
+          emptyDir:
+            medium: Memory
+            sizeLimit: "10Gi"
+        {{- end }}
+        {{- if .Values.persistence.hostPath.enabled }}
+        # Model weights are saved to host storage so they survive pod restarts on the same node.
+        - name: models
+          hostPath:
+            path: {{ .Values.persistence.hostPath.path }}/models
```
castabar: nit: When hostPath.enabled: true and path: "", the template renders:
```yaml
hostPath:
  path: /models
  type: DirectoryOrCreate
```

According to Kimi this would mount `/models` on the root filesystem instead of whatever volume mount the node has available.

I'd add a required guard as such
```yaml
    path: {{ required "persistence.hostPath.path must be set when hostPath is enabled" .Values.persistence.hostPath.path }}/models
```

castabar (approved):
one nit, otherwise lgtm

pmateusz (approved):

PR #1495, opened 2026-06-30
[castai-evictor] Evictor config crd with upgrade job.

jozuenoon:
The better evictor CRD config, no hooks.

---
### What changed
Introduces the `EvictorConfig` Custom Resource and CRD to the `castai-evictor` chart, mapping Helm values into a typed singleton CR. Adds a pre-install/pre-upgrade hook job to keep the CRD up to date and extends `customConfig` handling to support structured eviction rules alongside the legacy string format.

### Why
Centralizing evictor configuration in a cluster-scoped CR provides a single source of truth that can be reconciled by the CAST AI platform in managed mode or owned by the cluster operator in self-managed mode, without requiring application restarts for configuration changes.

### Key changes
- Added `EvictorConfig` CRD (`crds/evictorconfigs.autoscaling.cast.ai.yaml`) defining the `autoscaling.cast.ai/v1alpha1` `EvictorConfig` resource.
- Added `templates/evictorconfig.yaml` rendering a singleton `evictor-config` CR from values, including legacy mappings and new CR-pending fields.
- Added `templates/crd-upgrade-job.yaml`, a pre-install/pre-upgrade hook Job using `rancher/kubectl` to server-side apply the CRD.
- Extended `templates/_helpers.tpl` (`evictor.configMap.customConfig`) to support legacy string, structured slice, and map-with-`evictionRules` `customConfig` formats.
- Added new values in `values.yaml` for CR fields, `installCRDs`, and `crdUpgrade`.
- Added CI value files (`ci/test-values-object.yaml`, `ci/test-values-structured.yaml`) and unit tests (`tests/configmap_test.yaml`, `tests/evictorconfig_test.yaml`).
- Updated RBAC comment in `child-charts/castai-evictor-ext/templates/clusterrole.yaml`.
- Removed empty `templates/NOTES.txt`.
- Bumped chart version to `0.35.102`.

### Impact
- New CR-pending fields are rendered into the CR but not yet read by the evictor binary; they have no operational effect until a future evictor release consumes the CR.
- Legacy `customConfig` strings remain backward compatible. Structured `customConfig` now populates both the ConfigMap and the CR's `evictionRules`.
- The CRD upgrade hook requires permission to create/patch `customresourcedefinitions` and runs a Job during each `helm install`/`upgrade`.
- `EvictorConfig` is a singleton named `evictor-config`; edits in `castai` management mode are overwritten by the platform on snapshot sync.


oanatmaria (approved):

charts/castai-evictor/templates/evictorconfig.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: autoscaling.cast.ai/v1alpha1
+kind: EvictorConfig
+metadata:
+  name: evictor-config
```
oleksandrsan, line 4: Just be aware that existing e2e clusters might have this config from the previous version unclenaed, so the installation MIGHT fail with `Error: UPGRADE FAILED: unable to continue with update: EvictorConfig "evictor-config" in namespace "castai-agent" exists and cannot be imported into the current release: invalid ownership metadata` 
But this is expected and we will need to do one CRD cleanup in this case.

oleksandrsan (approved):

PR #1504, opened 2026-07-01
[castai-evictor] Evictor config CRD with upgrade job.

jozuenoon:
What changed

 Adds the EvictorConfig CRD and CR to the evictor chart with support for both standalone and umbrella chart installation in a single helm install.

 Why

 Moves evictor configuration into a typed Kubernetes custom resource that the evictor can watch and reconcile without pod restarts.

 Key changes
 - CRD definition lives in crd-files/ (not crds/) so Helm doesn't attempt native CRD install, which requires apiextensions.k8s.io permissions the
   cluster-controller SA lacks
 - crd-upgrade-job.yaml: pre-install/pre-upgrade hook Job that applies the CRD via kubectl apply --server-side
 - evictorconfig.yaml: renders as a regular template when the CRD already exists, or as a post-install hook on first install — solving the Helm chicken-and-egg
   problem where CRD and CR need to be created in one release
 - Guard uses .Capabilities.APIVersions.Has "autoscaling.cast.ai/v1alpha1/EvictorConfig" to check for the specific CRD,
 - Hook-created CR includes meta.helm.sh/release-* annotations for clean uninstall/re-install cycles
 - All existing Helm values are rendered into the CR; customConfig supports both legacy string and structured evictionRules formats

 Install flows
 - crdUpgrade.enabled=true (default), first install: pre-install job creates CRD → post-install hook creates CR
 - crdUpgrade.enabled=true, upgrade: job updates CRD schema → CR updated as regular template
 - crdUpgrade.enabled=false: no CRD-related resources, safe for clusters without CRD permissions


charts/castai-evictor/Chart.yaml
```diff
@@ -14,7 +14,7 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
-version: 0.35.117
+version: 0.35.118
```
primoz-hrovat: Will the auto-upgrade work properly?

jozuenoon: We are now sending the config from upgrade job in autoscaler to disable the crd update job and CR creation attempt.

nkotikov-castai:
LGTM, but I would've not trusted myself and wait for one more extra =)

oanatmaria (approved):

PR #1507, opened 2026-07-02
[castai-db-proxy] Sync connection draining with k8s termination

alijared:
---
### What changed
Adds configurable connection draining to the `castai-db-proxy` Helm chart. Pod `terminationGracePeriodSeconds` is now derived from the new draining settings, and the proxy configuration includes a `[server]` block with grace period values.

### Why
Enables controlled graceful shutdown of existing database connections during pod termination, aligning Kubernetes pod lifecycle configuration with the proxy's internal draining behavior.

### Key changes
- `values.yaml`: Added `connection_draining` block with `grace_period_seconds` (default `55`) and `graceful_shutdown_timeout_seconds` (default `5`).
- `templates/configmap.yaml`: Added `[server]` section populated from `connection_draining` values.
- `templates/deployment.yaml`: Added validation requiring both draining values to be positive integers; `terminationGracePeriodSeconds` is now computed as `grace_period_seconds + graceful_shutdown_timeout_seconds + 5`.
- `templates/_versions.tpl`: Bumped default proxy image version from `v0.2.0` to `v0.3.0`.
- `Chart.yaml` and `README.md`: Bumped chart version to `0.3.0` and documented new values.

### Impact
- The default proxy image tag is updated to `v0.3.0`; ensure the new image supports the `[server]` configuration block before deploying.
- `terminationGracePeriodSeconds` now defaults to `65` (`55 + 5 + 5`) instead of the previous hard-coded `60`.
- Invalid `connection_draining` values (zero or negative) will cause `helm install/upgrade` to fail with a validation error.

### DIFF
--- a/charts/castai-db-proxy/templates/deployment.yaml
+++ b/charts/castai-db-proxy/templates/deployment.yaml
@@ -33,7 +33,10 @@ spec:
       {{- if .Values.serviceAccountName }}
       serviceAccountName: {{ .Values.serviceAccountName }}
       {{- end }}
-      terminationGracePeriodSeconds: 60
+      {{- if or (lt (int .Values.connection_draining.grace_period_seconds) 1) (lt (int .Values.connection_draining.graceful_shutdown_timeout_seconds) 1) }}
+        {{- fail "connection_draining.grace_period_seconds and graceful_shutdown_timeout_seconds must be positive integers" }}
+      {{- end }}
+      terminationGracePeriodSeconds: {{ add .Values.connection_draining.grace_period_seconds .Values.connection_draining.graceful_shutdown_timeout_seconds 5 }}
       {{- with .Values.dnsPolicy }}
       dnsPolicy: {{ . }}
       {{- end }}
--- a/charts/castai-db-proxy/values.yaml
+++ b/charts/castai-db-proxy/values.yaml
@@ -130,3 +130,10 @@ dnsPolicy: ""
 
 # -- Pod DNS configuration.
 dnsConfig: {}
+
+# -- Connection draining configuration
+connection_draining:
+  # -- How long existing connections have to finish after SIGTERM. Must be > 0
+  grace_period_seconds: 55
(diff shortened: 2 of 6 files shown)


charts/castai-db-proxy/values.yaml
```diff
@@ -130,3 +130,10 @@ dnsPolicy: ""
 
 # -- Pod DNS configuration.
 dnsConfig: {}
```
viktorasm, line 132: is this used?

viktorasm (approved):

charts/castai-db-proxy/values.yaml
alijared: Yeah in the `deployment.yaml`

charts/castai-db-proxy/templates/deployment.yaml
```diff
@@ -33,7 +33,10 @@ spec:
       {{- if .Values.serviceAccountName }}
       serviceAccountName: {{ .Values.serviceAccountName }}
       {{- end }}
-      terminationGracePeriodSeconds: 60
+      {{- if or (lt (int .Values.connection_draining.grace_period_seconds) 1) (lt (int .Values.connection_draining.graceful_shutdown_timeout_seconds) 1) }}
```
viktorasm, line 36: nitpick: could this be optional instead? with changes like these in the future you'd be making upgrades to helm charts breaking for existing installations (introduces new mandatory fields that are not absolutely required)

alijared: The `[server]` config option itself has always been available, Kimchi's notes are wrong. The config for the graceful windows as far as the proxy is concerned have always existed because they're Pingora options, not options we defined. As for the deployment itself, `terminationGracePeriodSeconds` also has a default (`30s`), we're just setting it manually here to align with the Pingora settings so k8s doesn't send a `SIGKILL` while proxy is shutting down. If you set them to 0 or a negative number, Pingora itself will panic, so it needs strict enforcement here to be above that value otherwise proxy won't start because of Pingora

PR #1510, opened 2026-07-06
[castai-hosted-model] Enable tracing

igor-susic1:


castabar (approved):

charts/castai-hosted-model/child-charts/sglang/templates/deployment.yaml
```diff
@@ -73,9 +73,25 @@ spec:
             {{- range .Values.extraArgs }}
             - {{ . | quote }}
             {{- end }}
+            {{- if .Values.tracing.enabled }}
+            - "--enable-trace"
+            - "--otlp-traces-endpoint={{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}"
```
castabar, line 78: 🔴 **CRITICAL — SGLang will crash at startup with the default config.**

SGLang's `get_otlp_span_exporter()` passes the raw endpoint value directly to the OTel Python SDK's `HTTPSpanExporter`:

```python
# SGLang python/sglang/srt/observability/trace.py
def get_otlp_span_exporter(endpoint):
    protocol = os.environ.get(OTEL_EXPORTER_OTLP_TRACES_PROTOCOL, "grpc")
    ...
    elif protocol == "http/protobuf":
        return HTTPSpanExporter(endpoint=endpoint)  # ← raw "host:4318"
```

The `HTTPSpanExporter` expects a full URL like `http://host:4318/v1/traces`. Passing `host:4318` causes the `requests` library to interpret `host` as the URL scheme, resulting in an `InvalidSchema` error. SGLang wraps this in a `RuntimeError`:

```python
raise RuntimeError(f"initialize opentelemetry error:{e}. Please set correct otlp endpoint.")
```

Since the default `tracing.protocol` is `"http/protobuf"` and the default `tracing.otlpEndpoint` is `"observability-otel-gateway-opentelemetry-collector.tools.svc.cluster.local:4318"`, enabling tracing with defaults will prevent the server from starting.

The SGLang docs confirm the correct format: *"To use the HTTP/protobuf span exporter... point to an HTTP endpoint, for example, `http://0.0.0.0:4318/v1/traces`."*

**Fix:** Construct the URL based on protocol:
```yaml
{{- if eq .Values.tracing.protocol "http/protobuf" }}
- "--otlp-traces-endpoint=http://{{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}/v1/traces"
{{- else }}
- "--otlp-traces-endpoint={{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}"
{{- end }}
```

charts/castai-hosted-model/child-charts/sglang/templates/router-deployment.yaml
```diff
@@ -123,6 +123,23 @@ spec:
             {{- range .Values.router.extraArgs }}
             - {{ . | quote }}
             {{- end }}
+            {{- if .Values.tracing.enabled }}
+            - "--enable-trace"
+            - "--otlp-traces-endpoint={{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}"
```
castabar, line 128: 🔴 **CRITICAL — Same endpoint format issue as the worker template (see comment on `deployment.yaml`).** The router will also crash at startup with the default `http/protobuf` + `host:4318` config. The same fix (construct the full URL for HTTP/protobuf) is needed here.

charts/castai-hosted-model/child-charts/sglang/templates/deployment.yaml
```diff
@@ -73,9 +73,25 @@ spec:
...
+            {{- end }}
           env:
             - name: HF_HOME # Co-locate the HuggingFace cache with the download dir
               value: /models
+            {{- if .Values.tracing.enabled }}
+            {{- if .Values.tracing.protocol }}
+            - name: OTEL_EXPORTER_OTLP_TRACES_PROTOCOL
+              value: {{ .Values.tracing.protocol | quote }}
+            {{- end }}
+            - name: OTEL_SERVICE_NAME
```
castabar: 🟡 **Non-blocking — `OTEL_SERVICE_NAME` is overridden by SGLang and has no effect.**

SGLang's `process_tracing_init()` explicitly sets the service name in the OTel Resource:

```python
resource = Resource.create(
    attributes={
        SERVICE_NAME: server_name,  # ← explicit attribute overrides OTEL_SERVICE_NAME env var
    }
)
```

Where `server_name` is `--served-model-name` for the worker. The explicit attribute in `Resource.create()` takes precedence over the `OTEL_SERVICE_NAME` environment variable per OTel SDK semantics.

This means `tracing.serviceName` and this env var are dead code — the worker always reports `--served-model-name` as the service name regardless of what you set here. Not harmful, but misleading. Consider removing to avoid confusion, or verify if SGLang v0.5.13 handles this differently.

charts/castai-hosted-model/child-charts/sglang/templates/router-deployment.yaml
```diff
@@ -123,6 +123,23 @@ spec:
...
+            {{- end }}
+          {{- if .Values.tracing.enabled }}
+          env:
+            {{- if .Values.tracing.protocol }}
+            - name: OTEL_EXPORTER_OTLP_TRACES_PROTOCOL
+              value: {{ .Values.tracing.protocol | quote }}
+            {{- end }}
+            - name: OTEL_SERVICE_NAME
+              value: {{ printf "%s-router" (.Values.tracing.serviceName | default (.Values.model.servedName | default .Values.model.name)) | quote }}
```
castabar: 🟡 **Non-blocking — the `%s-router` suffix never appears in traces.**

Same issue as the worker: SGLang's `process_tracing_init()` sets `SERVICE_NAME` explicitly in the Resource, overriding `OTEL_SERVICE_NAME`. For the router, the server name is hardcoded to `sgl-router` inside SGLang's router launch code, so the `printf "%s-router"` value computed here has no effect on the actual service name reported in traces.

The `OTEL_SERVICE_NAME` env var and `tracing.serviceName` value are both dead code for the router.

charts/castai-hosted-model/child-charts/sglang/values.yaml
```diff
@@ -194,6 +194,27 @@ livenessProbe:
   httpGet:
     path: /health
 
+# -- End-to-end inference tracing. When enabled, both the worker
+# (sglang.launch_server) and, if deployed, the router (sglang_router.launch_router)
+# are launched with --enable-trace and export OTLP spans to otlpEndpoint.
+# NOTE: SGLang does not continue an upstream W3C traceparent — it starts its own
```
castabar: 🟡 **Non-blocking — this note may be inaccurate for recent SGLang versions.**

SGLang's current source (`python/sglang/srt/observability/trace.py`) uses `TraceContextTextMapPropagator` to extract `traceparent`/`tracestate` headers from incoming requests:

```python
TRACE_HEADERS = ["traceparent", "tracestate"]
...
external_trace_context = _trace_context_propagator.extract(
    self.external_trace_header or {}
)
root_span = tracer.start_span(
    name=..., context=external_trace_context, ...
)
```

And the SGLang router docs state: *"W3C Trace Context propagation for HTTP and gRPC"* and *"Trace context injection into upstream worker requests."*

This suggests SGLang **does** continue upstream traces when the `traceparent` header is present. Worth verifying against v0.5.13 — if it does support propagation, this note should be corrected as it may mislead users into thinking they can't get end-to-end traces.

castabar (commented):
## Review Summary

Adds OTLP tracing support to the SGLang child chart — both worker and router get `--enable-trace` + `--otlp-traces-endpoint` flags and OTel env vars. Clean implementation with good documentation. However, there is a **blocking issue** with the default endpoint format.

### 🔴 CRITICAL — SGLang will crash at startup with the default config

The default `tracing.protocol` is `"http/protobuf"` and the default `tracing.otlpEndpoint` is `"host:4318"`. SGLang's `get_otlp_span_exporter()` passes the raw endpoint directly to the OTel Python SDK's `HTTPSpanExporter`, which expects a full URL like `http://host:4318/v1/traces`. Passing `host:4318` causes an `InvalidSchema` error that SGLang wraps in a `RuntimeError` and does not catch — the server will fail to start.

**Fix:** Construct the URL based on protocol in both templates:
```yaml
{{- if eq .Values.tracing.protocol "http/protobuf" }}
- "--otlp-traces-endpoint=http://{{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}/v1/traces"
{{- else }}
- "--otlp-traces-endpoint={{ required "tracing.otlpEndpoint must be set when tracing.enabled is true" .Values.tracing.otlpEndpoint }}"
{{- end }}
```

### 🟡 Non-blocking findings

1. **`OTEL_SERVICE_NAME` env var is overridden by SGLang** — SGLang's `process_tracing_init()` explicitly sets `SERVICE_NAME` in the OTel `Resource.create()` call (using `--served-model-name` for the worker, hardcoded `sgl-router` for the router). The explicit resource attribute takes precedence over the `OTEL_SERVICE_NAME` env var, so `tracing.serviceName` has no effect. The router's `printf "%s-router"` suffix never appears in traces.

2. **W3C traceparent note may be inaccurate** — SGLang's current source uses `TraceContextTextMapPropagator` to extract `traceparent`/`tracestate` from incoming request headers and passes the extracted context to `tracer.start_span()`. The router docs also mention "W3C Trace Context propagation." Worth verifying against v0.5.13 — SGLang may actually continue upstream traces.

### ✅ What's good
- Clean conditional rendering, tracing disabled by default
- Proper `required` validation for `otlpEndpoint`
- Correct handling of `SGLANG_TRACE_LEVEL` with `| toString | ne ""` (avoids `0` being falsy)
- Chart.lock updated with correct digest
- Version bumps appropriate (minor for feature, patch for parent dependency)

castabar (changes requested):

igor-susic1:
@getkimchi review

PR #1519, opened 2026-07-15
[castai-hosted-model] add protection to the router deployments (PDB,…

igor-susic1:
PDB, pod anti-affinity, zero-downtime single-replica rollout

Adds availability/resilience controls to the SGLang router in the castai-hosted-model chart.

What's included:
- PodDisruptionBudget for the router (router.pdb, default off) — blocks voluntary evictions (node drain, CAST scale-down, node-pool rollout) from taking down the router. Supports minAvailable/maxUnavailable (mutually exclusive, validated).
- Pod anti-affinity (router.podAntiAffinity, default on) — auto-spreads replicas one-per-node when replicaCount > 1, so a single node loss can't take out the whole router set. Merges with any existing affinity; required (hard) by default.
- Zero-downtime single-replica rollout (router.zeroDowntimeSingleReplica, default on) — at replicaCount: 1, uses RollingUpdate surge (new router Ready before old is torn down) instead of Recreate, eliminating the serving gap on deploy.

Version bumps: sglang child chart 0.1.2 → 0.2.0; parent castai-hosted-model 0.0.59 → 0.0.60 (+ dep pin + Chart.lock).

Scope: chart-only; all features backward-compatible (PDB off by default).


charts/castai-hosted-model/child-charts/sglang/templates/router-deployment.yaml
```diff
@@ -54,8 +54,20 @@ spec:
       maxSurge: {{ .Values.router.strategy.rollingUpdate.maxSurge }}
       maxUnavailable: {{ .Values.router.strategy.rollingUpdate.maxUnavailable }}
     {{- end }}
+    {{- else if .Values.router.zeroDowntimeSingleReplica }}
```
Zilv1nas: Not sure why are we making this so complicated, the chart could just allow passing rolling update settings and deployment strategy, similar to e.g. vLLM deployment settings

Zilv1nas (approved):

PR #1533, opened 2026-07-28
[castai-evictor] CRD upgrade job pod labels configuration.

jozuenoon:
Manage pod labels and annotations for the crd upgrade job separately.

### DIFF
--- a/charts/castai-evictor/templates/crd-upgrade-job.yaml
+++ b/charts/castai-evictor/templates/crd-upgrade-job.yaml
@@ -99,7 +99,14 @@ spec:
     metadata:
       name: {{ include "evictor.fullname" . }}-crd-upgrade
       labels:
-        {{- include "evictor.labels" . | nindent 8 }}
+        {{- include "evictor.selectorLabels" . | nindent 8 }}
+        {{- with .Values.crdUpgrade.podLabels }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.crdUpgrade.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
       restartPolicy: Never
       serviceAccountName: {{ include "evictor.fullname" . }}-crd-upgrade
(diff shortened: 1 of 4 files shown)


charts/castai-evictor/templates/crd-upgrade-job.yaml
```diff
@@ -93,13 +93,21 @@ metadata:
...
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    {{- include "evictor.annotations" . | nindent 4 }}
 spec:
   backoffLimit: 3
   template:
     metadata:
       name: {{ include "evictor.fullname" . }}-crd-upgrade
       labels:
-        {{- include "evictor.labels" . | nindent 8 }}
+        {{- include "evictor.selectorLabels" . | nindent 8 }}
```
primoz-hrovat: Why was this changed?

charts/castai-evictor/templates/crd-upgrade-job.yaml
```diff
@@ -99,7 +99,14 @@ spec:
...
+        {{- with .Values.crdUpgrade.podLabels }}
```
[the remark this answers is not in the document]
jozuenoon: The change aligns the Job pod template with the existing Deployment pod template.

charts/castai-evictor/templates/crd-upgrade-job.yaml
```diff
@@ -99,7 +99,14 @@ spec:
     metadata:
```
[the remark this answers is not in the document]
jozuenoon: This can't happen here — two mechanisms prevent it:

 1. The Job name includes {{ .Release.Revision }} (e.g., castai-evictor-crd-upgrade-1, castai-evictor-crd-upgrade-2), so every helm upgrade creates a new Job rather than updating the
    existing one. Selector immutability only applies to in-place updates of the same Job.

 2. The hook has helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded, so the previous Job is always deleted before the new one is created.

 No selector conflict is possible regardless of what the user puts in crdUpgrade.podLabels.

charts/castai-evictor/templates/crd-upgrade-job.yaml
jozuenoon: Aligned to what is in the Deployment, the `evictor.labels` are defined in helpers and should be in the metadata, in pod labels the selectors are sufficient.

jozuenoon: But actually we may also even drop that in favor for custom labels for the job.

jozuenoon:
cc @j-furtado

primoz-hrovat (approved):

PR #1534, opened 2026-07-28
[castai-evictor] Split CRD/CR apply into initcontainers.

jozuenoon:
1. Split the CRD and CR creation into initcontainers (single job while ensuring the proper sequence).
2. Add security context for crdUpgrade and OpenshiftSCC handling based on the Evictor deployment.


charts/castai-evictor/tests/crd-upgrade-job_test.yaml
```diff
@@ -0,0 +1,154 @@
+suite: CRD upgrade Job rendering
```
primoz-hrovat, line 1: I don't see the runner (action) to actually run these tests.

PR #1549, opened 2026-08-07
CSU-5760: Replace Docker Hub alpine/k8s default with CAST AI mirror in chart-upgrader and gpu-metrics-exporter

0msokolowski0:


plks (approved):

charts/castai-chart-upgrader/values.yaml
```diff
@@ -62,7 +62,7 @@ rbac:
 
 # Container image for the upgrade job.
 image:
-  repository: "alpine/k8s"
+  repository: "us-docker.pkg.dev/castai-hub/library/alpine/k8s"
```
radekska, line 65: how do we authenticate against the registry? or is it publicly available

0msokolowski0: It's publicly available, the version used here is also mirrored.

0msokolowski0: It's publicly available, the version used here is also mirrored.

0msokolowski0: This PR was split into 2, forgot to close this one.

PR #1570, opened 2026-08-20
Fix umbrella release job

oanatmaria:


.github/workflows/update-umbrella-on-child-release.yaml
```diff
@@ -45,26 +45,16 @@ jobs:
           echo "skip=false" >> $GITHUB_OUTPUT
           echo "Parsed chart: ${CHART_NAME}, version: ${VERSION}"
```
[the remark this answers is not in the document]
oanatmaria: by the time the tag is pushed to github, the helm chart was already released by the gitlab pipelines so we don't need to wait anymore

.github/workflows/update-umbrella-on-child-release.yaml
```diff
@@ -45,26 +45,16 @@ jobs:
...
-          allowed-conclusions: success
-
   update-umbrella:
-    needs: wait-for-release
+    needs: parse-release-tag
     if: |
       !cancelled() &&
       (github.event_name == 'workflow_dispatch' ||
-       (needs.wait-for-release.result == 'success' && needs.wait-for-release.outputs.skip != 'true'))
+       (needs.parse-release-tag.result == 'success' && needs.parse-release-tag.outputs.skip != 'true'))
```
oskarwojciski, line 53: Is it still needed? Maybe we can just leave option to manually dispatch the full workflow?

oskarwojciski (approved):

.github/workflows/update-umbrella-on-child-release.yaml
oanatmaria: we discuss on Discord that we will keep it as it is for now

PR #1582, opened 2026-08-28
draft: [castai-db-proxy] Adding ProxySQL and enable pooling for MySQL

balazskvancz:


charts/castai-db-proxy/values.yaml
```diff
@@ -226,3 +229,52 @@ pooling:
...
+    # -- Enable ProxySQL pooler sidecar (MySQL only)
+    enabled: false
+    image:
+      repository: docker.io/proxysql/proxysql
+      pullPolicy: IfNotPresent
+      tag: ""
+    resources:
+      cpu: "500m"
+      memoryRequest: "256Mi"
+      memoryLimit: "256Mi"
```
alijared: I'd use the same resources as PgDog by default

alijared (approved):

charts/castai-db-proxy/values.yaml
```diff
@@ -226,3 +229,52 @@ pooling:
...
+      # -- Backend connection timeout in milliseconds
+      connect_timeout_server: 5000
+      # -- Default query timeout in milliseconds
+      default_query_timeout: 36000000
+      # -- Delay multiplexing for N queries after auto-increment INSERT
+      auto_increment_delay_multiplex: 5
+      # -- Stacksize for ProxySQL threads
+      stacksize: 1048576
+      # -- Enable have_compress
+      have_compress: true
```
westwoodra2: We are duplicating the default values -- once here in `values.yaml` and once in `templates/proxysql-config.yaml`, if I look at PgDog it has it's config empty i.e. `config: {}` in this file.

charts/castai-db-proxy/templates/proxysql-config.yaml
```diff
@@ -0,0 +1,79 @@
...
+      {
+        rule_id={{ add $index 1 }}
+        active=1
+        destination_hostgroup={{ ternary 1 0 .readonly }}
+        apply=1
+        multiplex={{ if $mergedConfig.multiplexing }}2{{ else }}0{{ end }}
+      }{{ if lt $index (sub (len $.Values.endpoints) 1) }},{{ end }}
+    {{- end }}
+    )
+{{- end }}
```
westwoodra2, line 80: ProxySQL rules usually have `match_pattern` to decide where the query goes. In this case we don't include any, so the first rule that will be generated will capture all the traffic and will send it to whatever `destination_hostgroup` will be specified. For example, if `destination_hostgroup=0` lands in the file then all reads and writes will go to write server and none to the read replicase.

ProxySQL docs [warn about this here](https://proxysql.com/documentation/main-runtime/mysql-tables/#:~:text=A%20general%20point%20to%20keep%20in%20mind%3A%20it%20is%20important%20NOT%20to%20set%20either%20the%20destination_hostgroup%20or%20apply%20fields%20in%20the%20first%20rule%20that%20captures%20all%20incoming%20traffic%2C%20as%20this%20will%20prevent%20any%20other%20rules%20from%20being%20evaluated.).

westwoodra2: We can fix it by adding a `proxy_port` into the rules as described in this [ProxySQL documentation](https://proxysql.com/documentation/proxysql-read-write-split-howto/#:~:text=PROXYSQL%20RESTART%3B-,Add%20routing%20rules%3A,-INSERT%20INTO%20mysql_query_rules).

So we will have smth like this:

```
mysql_query_rules=
(
  { rule_id=1  active=1  proxy_port=6033  destination_hostgroup=0  apply=1 },
  { rule_id=2  active=1  proxy_port=6034  destination_hostgroup=1  apply=1 }
)
```

It should work cause proxy itself will send write queries to 6033 and read queries to 6034.

westwoodra2: Maybe smth like this:

The `hasReadonly` guard first:
```
{{- $hasReadonly := false -}}
{{- range .Values.endpoints }}{{- if .readonly }}{{- $hasReadonly = true }}{{- end }}{{- end }}
```

and the rules themselves:

```
 mysql_query_rules=
    (
      {
        rule_id=1
        active=1
        proxy_port={{ .Values.pooling.proxySql.readWritePort }}
        destination_hostgroup=0
        apply=1
        multiplex={{ if $mergedConfig.multiplexing }}2{{ else }}0{{ end }}
      }
      {{ if $hasReadonly }},
      {
        rule_id=2
        active=1
        proxy_port={{ .Values.pooling.proxySql.readonlyPort }}
        destination_hostgroup=1
        apply=1
        multiplex={{ if $mergedConfig.multiplexing }}2{{ else }}0{{ end }}
      }{{ end }}
    )
  ```

File: charts/castai-db-proxy/templates/pooling-deployment.yaml [hunk 4]
```diff
@@ -54,6 +89,81 @@ spec:
...
+
+                while IFS= read -r line || [ -n "$line" ]; do
+                  case "$line" in ''|\#*) continue ;; esac
+                  case "$line" in *:*) ;; *) echo "invalid line (expected username:password): $line" >&2; exit 1 ;; esac
+                  user="${line%%:*}"
+                  pass="${line#*:}"
+                  [ "$first" -eq 1 ] && first=0 || echo ","
+                  printf '  { username="%s" , password="%s" , default_hostgroup=%s , active=1 }' "$user" "$pass" "$hostgroup"
+
+                  hostgroup=$((hostgroup + 1))
```
westwoodra2, line 159: We should leave `default_hostgroup=0` for all users. ProxySQL docs say that when a query comes from a user and no `mysql_rules` can match it, then the `default_hostgroup` will be used to send the query to the server. So it theoretically can be that a user's write query can go to a read replica if user's `default_group` will be set to `1`.

charts/castai-db-proxy/templates/pooling-deployment.yaml
```diff
@@ -54,6 +89,81 @@ spec:
...
+                first=1
+                hostgroup=0
+
+                while IFS= read -r line || [ -n "$line" ]; do
+                  case "$line" in ''|\#*) continue ;; esac
+                  case "$line" in *:*) ;; *) echo "invalid line (expected username:password): $line" >&2; exit 1 ;; esac
+                  user="${line%%:*}"
+                  pass="${line#*:}"
+                  [ "$first" -eq 1 ] && first=0 || echo ","
+                  printf '  { username="%s" , password="%s" , default_hostgroup=%s , active=1 }' "$user" "$pass" "$hostgroup"
```
westwoodra2: We should escape the password just in case -- for that we can use `%q` instead of `%s`

charts/castai-db-proxy/templates/pgdog-config.yaml
```diff
@@ -0,0 +1,63 @@
+{{- if and .Values.pooling.enabled .Values.pooling.pgdog .Values.pooling.pgdog.enabled }}
```
westwoodra2: This PR introduced `pooling.pgdog.enabled` and `pooling.proxysql.enabled`, but I don't see in the follow-up PR that anyone sets those values. Is this gonna be 3rd follow-up PR?

westwoodra2 (changes requested):
We need to fix some things first.

PR #1605, opened 2026-09-14
[castai-db-proxy] adding ProxySQL support for MySQL pooling

balazskvancz:


charts/castai-db-proxy/templates/pooling-configmap.yaml
```diff
@@ -1,82 +1,41 @@
 {{- if .Values.pooling.enabled }}
+{{- $pgdogEnabled := and .Values.pooling.pgdog  (eq .Values.protocol "PostgreSQL") -}}
+{{- $proxySqlEnabled := and .Values.pooling.proxySql (eq .Values.protocol "MySQL") -}}
+{{- if and $pgdogEnabled $proxySqlEnabled }}
+{{- fail "pooling.pgdog and pooling.proxySql are mutually exclusive; enable only one" }}
+{{- end }}
```
viktorasm: will never happen as protocol can only be one or the other, both will never be true

charts/castai-db-proxy/templates/pooling-configmap.yaml
```diff
@@ -1,82 +1,41 @@
...
+{{- if and $proxySqlEnabled (ne .Values.protocol "MySQL") }}
+{{- fail "pooling.proxySql requires protocol: \"MySQL\"" }}
+{{- end }}
```
viktorasm, line 6: do we have similar guard for pgdog+!postgres combo?

charts/castai-db-proxy/templates/pooling-configmap.yaml
```diff
@@ -1,82 +1,41 @@
...
     url = "postgresql://{{ .Values.pooling.pgdog.admin.user | urlquery }}:{{ .Values.pooling.pgdog.admin.password | urlquery }}@localhost:{{ .Values.pooling.pgdog.port }}/{{ .Values.pooling.pgdog.admin.name | urlquery }}"
     {{- end }}
 
+    {{- if $proxySqlEnabled }}
+    [pooler.proxy_sql]
+    metrics_url = "http://localhost:{{ .Values.pooling.metricsPort }}/metrics"
+    {{- end }}
+
     [api]
     url = {{ printf "https://%s" (required "apiURL must be provided" .Values.apiURL) | quote }}
```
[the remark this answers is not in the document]
viktorasm: i think this is a valid comment

charts/castai-db-proxy/templates/pooling-deployment.yaml
```diff
@@ -54,6 +75,81 @@ spec:
...
+                - ALL
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+            runAsUser: 65532
+        - name: proxysql-users-init
+          image: "{{ .Values.pooling.proxySql.image.repository }}:{{ include "castai-db-proxy.proxySqlImage" . }}"
+          imagePullPolicy: {{ .Values.pooling.proxySql.image.pullPolicy }}
+          envFrom:
+            - secretRef:
+                name: {{ required "userSecretRef must be provided" .Values.pooling.proxySql.userSecretRef | quote }}
```
alijared: To make it clearer what the env vars are actually meant to be called, I'd rather have something like this because it also clearly illustrates what/how the secret itself is meant to be defined 

```suggestion
          env:
            - name: USERNAME
              valueFrom:
                secretKeyRef:
                  name: {{ required "userSecretRef must be provided" .Values.pooling.proxySql.userSecretRef | quote }}
                  key: DATABASE_USERNAME
            - name: PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ required "userSecretRef must be provided" .Values.pooling.proxySql.userSecretRef | quote }}
                  key: DATABASE_PASSWORD
```

alijared (changes requested):

charts/castai-db-proxy/templates/pooling-deployment.yaml
```diff
@@ -54,6 +89,132 @@ spec:
             readOnlyRootFilesystem: true
             runAsNonRoot: true
             runAsUser: 65532
+        {{- end }}
+        {{- if $proxySqlEnabled }}
+        - name: proxysql-init
+          image: "{{ .Values.pooling.proxySql.image.repository }}:{{ include "castai-db-proxy.proxySqlImage" . }}"
```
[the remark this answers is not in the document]
viktorasm: might be valid, might fail outside macos

charts/castai-db-proxy/templates/pooling-deployment.yaml
viktorasm: +1, wanted to comment the same

charts/castai-db-proxy/templates/pooling-configmap.yaml
balazskvancz: True, it remained here from previous implementation. Deleted.

charts/castai-db-proxy/templates/pooling-deployment.yaml
balazskvancz: Makes sense, implemented.

charts/castai-db-proxy/templates/pooling-configmap.yaml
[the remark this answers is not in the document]
balazskvancz: I think its a false positive comment, since the default value is already there.

charts/castai-db-proxy/templates/pooling-configmap.yaml
balazskvancz: Currently not, but I can add.

charts/castai-db-proxy/templates/pooling-deployment.yaml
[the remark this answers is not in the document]
balazskvancz: Done.

viktorasm (approved):

charts/castai-db-proxy/templates/pooling-configmap.yaml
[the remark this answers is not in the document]
viktorasm: disregard then

charts/castai-db-proxy/templates/pooling-configmap.yaml
viktorasm: nevermind - i think this is also covered by definition of proxysqlEnabled. this condition will never be true

alijared (approved):

PR #1621, opened 2026-09-23
[castai-umbrella] enable KENT auto upgrades

mindaugasCast:


zamai (approved):
Finally!!!

zamai (approved):
Finally!!!

